PR #5, opened 2020-07-08
Implemented user facing objects attributes change from pointers to no…

vidhutsingh:
Feature :
Implemented user facing objects attributes changes from pointers to non pointers with few exceptions . 
1. Other object type or list of object types are still pointers . Most of these attributes are output type and there are few input attributes. so, shouldn’t be major concern. 
2. Enums are still pointers .  However internally it is handled and user doesnt have to worry about creating pointer by their own. 
3. Fix formatting in the code base. 

Details :
Implemented encoding and decoding function in objectset layer for all the objectset types . The encoding and decoding basically converts the object from non pointers to pointer type and vice-versa. 

Test :
xxxxx:service xxxxx.xx$ go test -v
--- PASS: TestVolumeServiceTestSuite (9.63s)
    --- PASS: TestVolumeServiceTestSuite/TestGetNonExistentVolumeByID (9.63s)
PASS
ok  	github.com/hpe-storage/nimble-golang-sdk/pkg/service	9.647s


pkg/client/access_control_records_object_set.go
```diff
@@ -23,17 +23,17 @@ type AccessControlRecordObjectSet struct {
 // CreateObject creates a new AccessControlRecord object
 func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *model.AccessControlRecord) (*model.AccessControlRecord, error) {
 	newPayload, err := model.EncodeAccessControlRecord(payload)
-	resp, err := objectSet.Client.Post(accessControlRecordPath, newPayload)
+	accessControlRecordObjectSetResp, err := objectSet.Client.Post(accessControlRecordPath, newPayload)
```
rgcostea: Why not just use resp?  It's shorter :)

vidhutsingh: Sure :)

pkg/client/access_control_records_object_set.go
```diff
@@ -3,68 +3,69 @@
...
+	Client *GroupMgmtClient
 }
 
 // CreateObject creates a new AccessControlRecord object
-func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *model.AccessControlRecord) (*model.AccessControlRecord, error) {
-	accessControlRecordObjectSetResp, err := objectSet.Client.Post(accessControlRecordPath, payload)
-	if err !=nil {
-		return nil,err
+func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
+	newPayload, err := nimbleos.EncodeAccessControlRecord(payload)
```
rkumpf: What is the expected behavior if this err is non-nil?  The value is overwritten on the next line.

pkg/client/access_control_records_object_set.go
```diff
@@ -3,68 +3,69 @@
...
+	resp, err := objectSet.Client.Post(accessControlRecordPath, newPayload)
+	if err != nil {
+		return nil, err
 	}
-	
+
 	// null check
-	if accessControlRecordObjectSetResp == nil {
-		return nil,nil
+	if resp == nil {
```
rkumpf, line 31: What is the behavior of nimbleos.DecodeAccessControlRecord when given a nil value?  If it could simply return nil, we wouldn't need this condition.  That simplification seems preferable, since I guess the response shouldn't be nil when no error is returned.  This pattern is repeated in other places.

pkg/client/access_control_records_object_set.go
```diff
@@ -3,68 +3,69 @@
 package client
 
 import (
-	"reflect"
 	"fmt"
 	"github.com/hpe-storage/common-host-libs/jsonutil"
-	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/model"
-	"github.com/hpe-storage/nimble-golang-sdk/pkg/util"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
```
rkumpf, line 8: This patch is much too large to be reviewed effectively.  When making pervasive changes such as renaming packages, I recommend limiting the patch to contain only those changes.  Otherwise a reviewer can easily miss logic changes in all the noise.  It is particularly difficult to review the changes in the files that were relocated (since git failed to correlate them).

pkg/client/disks_object_set.go
```diff
@@ -46,21 +46,21 @@ func (objectSet *DiskObjectSet) DeleteObject(id string) error {
 }
 
 // GetObject returns a Disk object with the given ID
-func (objectSet *DiskObjectSet) GetObject(id string) (*model.Disk, error) {
-	diskObjectSetResp, err := objectSet.Client.Get(diskPath, id, model.Disk{})
+func (objectSet *DiskObjectSet) GetObject(id string) (*nimbleos.Disk, error) {
+	diskObjectSetResp, err := objectSet.Client.Get(diskPath, id, nimbleos.Disk{})
```
rkumpf: Use "resp" for consistency.

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -0,0 +1,133 @@
...
+	CreationTime *int64 `json:"creation_time,omitempty"`
+	// LastModified - Time when this access control record was last modified.
+	LastModified *int64 `json:"last_modified,omitempty"`
+	// AccessProtocol - Access protocol of the volume.
+	AccessProtocol *NsAccessProtocol `json:"access_protocol,omitempty"`
+}
+
+// EncodeAccessControlRecord - Transform AccessControlRecord to accessControlRecord type
+func EncodeAccessControlRecord(request interface{}) (*accessControlRecord, error) {
+	reqAccessControlRecord := request.(*AccessControlRecord)
```
rkumpf: We accept an argument of any type, and then panic if it is not the expected type.  Why not declare the parameter to be of the expected type?  Alternatively, we could check the "ok" value from the type assertion and return an error gracefully.

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -0,0 +1,133 @@
...
+	// CreationTime - Time when this access control record was created.
+	CreationTime int64 `json:"creation_time,omitempty"`
+	// LastModified - Time when this access control record was last modified.
+	LastModified int64 `json:"last_modified,omitempty"`
+	// AccessProtocol - Access protocol of the volume.
+	AccessProtocol *NsAccessProtocol `json:"access_protocol,omitempty"`
+}
+
+// sdk internal struct
+type accessControlRecord struct {
```
rkumpf: Can we differentiate this name from the upper case form?

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -0,0 +1,133 @@
...
+	// ID - Identifier for the access control record.
```
rkumpf: Is there a reason to duplicate all these comments?  Maybe we can just direct the reader to the other class for field descriptions.

pkg/client/access_control_records_object_set.go
```diff
@@ -3,68 +3,69 @@
...
 }
 
 // CreateObject creates a new AccessControlRecord object
-func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *model.AccessControlRecord) (*model.AccessControlRecord, error) {
-	accessControlRecordObjectSetResp, err := objectSet.Client.Post(accessControlRecordPath, payload)
-	if err !=nil {
-		return nil,err
+func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
+	newPayload, err := nimbleos.EncodeAccessControlRecord(payload)
+	resp, err := objectSet.Client.Post(accessControlRecordPath, newPayload)
```
rkumpf: The name "new" payload isn't very instructive. Maybe "encodedPayload"? Except it isn't really encoded, it is just transformed to a different object model. The word "encoded" makes me think it is serialized, such as in JSON.

Can we clarify the object models we are transforming between? Are the internal, lower-case structures based on an existing Nimble array interface, while the upper-case structures are our choice for how to provide this SDK? If so, perhaps "nimbleosPayload" would be appropriate.

Perhaps my comment further down will make this one moot.  :)

pkg/client/access_control_records_object_set.go
```diff
@@ -3,68 +3,69 @@
...
+
 	// null check
-	if accessControlRecordObjectSetResp == nil {
-		return nil,nil
+	if resp == nil {
+		return nil, nil
 	}
-	return accessControlRecordObjectSetResp.(*model.AccessControlRecord), err
+
+	return nimbleos.DecodeAccessControlRecord(resp)
```
rkumpf: Related to the comment above, could we make the Encode* and Decode* names more representative, such as ToNimble* and FromNimble*?

pkg/param/get_params.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+
+package param
```
rkumpf, line 3: I see that the util package is renamed to param, and I see that its files were copied to the param directory, but I don't see that the files were removed from the util directory.  Did I miss it?

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -0,0 +1,133 @@
...
+	byte, err := json.Marshal(reqAccessControlRecord)
+	objPtr := &accessControlRecord{}
+	err = json.Unmarshal(byte, objPtr)
+	return objPtr, err
+}
+
+// DecodeAccessControlRecord Transform accessControlRecord to AccessControlRecord type
+func DecodeAccessControlRecord(request interface{}) (*AccessControlRecord, error) {
+	reqAccessControlRecord := request.(*accessControlRecord)
+	byte, err := json.Marshal(reqAccessControlRecord)
```
rkumpf: Please check my understanding.  We currently have the model classes defined in such a way that they work well with the client   Now we are deciding that we want some minor changes to the way those model classes are exposed from the SDK.  To preserve compatibility with the client, we create a second model and transformations between them.  That way we can expose the SDK interfaces we want without affecting interaction with the Nimble devices.  To enable simple transformation between the models, we set them up such that their JSON encodings are identical.

However, the underlying REST client has no apparent dependency on the input model structure.  It just needs to be able to convert the payload to/from JSON.  So, I wonder if the internal, lower-case model classes aren't needed at all, and we can simply modify the existing model as we see fit as long as it continues to produce proper JSON.

rkumpf (commented):
I've added review comments, some of them perhaps significant, but I don't know that any of them strictly must be addressed before submittal.  If you agree, I suggest submitting this patch as is and addressing comments in subsequent (much, much smaller) patches.  Attempting further refinements in this giant patch would make it inordinately difficult for reviewers to catch mistakes.

rkumpf (commented):
How much effort would be required to put up a separate pull request with just the package rename changes, redone from scratch rather than as a modification from the current state of this pull request?  If not so much effort, I'd suggest that as a first step.  Then the remaining changes can be applied on top of that, and we can see exactly what they are.  Also, if you agree with my speculation that separate "user facing" objects might not be needed, it should greatly simplify the remaining changes.

pkg/client/access_control_records_object_set.go
vidhutsingh: ya, I need to fix this . We need to handle the error. thanks.

pkg/client/access_control_records_object_set.go
vidhutsingh: We cant do that as it will crash in the nimbleos.DecodeAccessControlRecord fn since it accepts interface as argument convert it to object type.

pkg/client/access_control_records_object_set.go
vidhutsingh: That is true but at some point we have to make this change and it will be always like this. If you need more context I can help you with that it will help you to follow the changeset. It looks huge because of multiple nimble objects. You just follow one object change i.e volume, flow would be volume_service.go (it uses volume.go)-> volumes_object_set.go -> group_mgmt_client.go . Similarly for all the objects.

pkg/client/disks_object_set.go
vidhutsingh: ok

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: No, there are 60+ different types of objects, and writing a type assertion for each of them doesn't add any value. Although, I do need to add err check in the Encoder and Decoder which I have missed.

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: Probably it is coming from the generator . we can fix that .

pkg/client/access_control_records_object_set.go
vidhutsingh: Actually, we are doing the encoding and decoding if you check the implementation. "encodedPayload" make sense. I will change it.

pkg/client/access_control_records_object_set.go
vidhutsingh: Please check above comment.

pkg/param/get_params.go
vidhutsingh: I'll check . thanks

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: Model classes are basically nimble objects. We cannot use non-pointers to attributes because sdk model class has both request/response attributes in the same class and we really dont want to send response attributes to request API call since the nimble array wont accept it. So we decided to use pointers, as nil pointers are dropped by json request. But that created another sdk user visible issue since now user has to create pointers to attributes for all the attributes of the object which is not very intuitive. Thus we introduced encoding/decoding which converts non-pointers to pointers. Hope this helps .

pkg/client/access_control_records_object_set.go
rkumpf: What format is it encoded into?  It seemed to me like it was being transformed to a different data model, not serialized into an encoding.  I suppose we are just quibbling over terminology.

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: I appreciate your patience as I come up to speed here.  I'm trying to understand why we don't simply declare the function as:

func EncodeAccessControlRecord(request *AccessControlRecord) (*accessControlRecord, error) {

Empty interfaces are used by code that handles values of unknown type.  It seems we know what type of value we're expecting to receive.

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: I see I'm missing a lot of history.  Thanks for sharing it with me.  I'm still unsure how we suppress the unwanted attributes, though.  Is an "unset" attribute marshalled in such a way that the unmarshalling recognizes it is unset and leaves the pointer nil?  I'd love to see an example.

Do we depend on the SDK user to leave the correct attributes unset?  How do they know which attributes are applicable to which actions?

pkg/client/access_control_records_object_set.go
rkumpf: I was suggesting we could change nimbleos.DecodeAccessControlRecord to be nil-safe.  Would that be problematic?

pkg/client/access_control_records_object_set.go
rkumpf: I agree that renaming the packages will be a big change however it is done, but it will be a purely mechanical change.  The difficulty arises when mechanical changes are intermingled with other changes, making the other changes hard to find.  I thought that's what is happening here, but perhaps I misunderstand.

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: This is a good starting point  https://www.sohamkamani.com/golang/2018-07-19-golang-omitempty/ link. That will be part of documentation. Also if you happen to set the incorrect attribute and nimos will return an error SM_unexpected_arg.

pkg/client/access_control_records_object_set.go
vidhutsingh: I am basically leveraging json marhsalling/unmarshalling to convert struct non-pointers attributes to pointers and vis-versa.  it is basically doing to conversion from data to json string type and vis-versa.

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: Thanks for the pointer.  I understand the idea of using pointers to indicate whether a value is set.  What I'm not following is how we get from the user model to a correct pointer value, nil or non-nil.  Let's suppose we have a user model class called Size that contains an int.  The internal "size" class will contain a *int.  If a "Size" object has a value of "0", will the "size" object contain nil or a pointer to 0?  How does it decide?  Maybe the user really is trying to update an attribute from some other value to 0.

pkg/client/access_control_records_object_set.go
rkumpf: Right, but the Encode method does both a marshal and unmarshal.  So does the Decode method.  The fact that we are using JSON as a convenience is an implementation detail, not an outwardly visible function.  The net result from the caller's perspective is a model transformation.

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: In EncodeAccessControlRecord() it can be done but again in DecodeAccessControlRecord() it cant be done. That will break the consistency . Since, during EncodeAccessControlRecord this will accepts concrete data type and during DecodeAccessControlRecord() an generic data type.  Moreover, we still need this type resp== nil check, it will just move to EncodeAccessControlRecord().

pkg/client/access_control_records_object_set.go
vidhutsingh: Please check the answer below .

pkg/client/access_control_records_object_set.go
vidhutsingh: Yes right.

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: The Encode and Decode functions look symmetric to me.  A function named "DecodeAccessControlRecord" would seem to be specific to the accessControlRecord type.  Why does it need to be generic?

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: I see your point, the value will be nil for size in your case.  Let me see how this can be handled.

pkg/client/v1/nimbleos/access_control_record.go
vidhutsingh: Because we dont know the POST return type until we land in the specific object set . The implementation of POST in group_mgmt_client.go is generic . please check .

rkumpf: See my comment at access_control_records_object_set.go line 29.

pkg/client/access_control_records_object_set.go
```diff
@@ -3,90 +3,95 @@
...
-func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *model.AccessControlRecord) (*model.AccessControlRecord, error) {
-	accessControlRecordObjectSetResp, err := objectSet.Client.Post(accessControlRecordPath, payload)
-	if err !=nil {
-		return nil,err
+func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
+	encodedPayload, err := nimbleos.EncodeAccessControlRecord(payload)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := objectSet.Client.Post(accessControlRecordPath, encodedPayload)
```
rkumpf: Subsequent code trusts that the "resp" value is an accessControlRecord structure.  Why?  Because the encodedPayload we pass in is an accessControlRecord structure, and Post returns the same type it is given.  (Note: This needs to be a documented behavior.)  By the same token, why wouldn't we trust that "resp" is not nil?  I think you are saying it is not practical to verify the response type but it is practical to verify a non-nil response.  I'd like to understand the distinction.

The Post function is able to return the correct type because it simply returns a filled in copy of the structure it is given.  (Or is it actually the same structure?)  I wonder if this could be problematic in some cases.  I suspect the JSON unmarshal logic leaves the existing structure contents in place and simply overwrites attributes as it decodes them.  Is that correct?  If so, it might be mishandling write-only attributes.  Consider a "set password" API.  The password value would be included in the request but not in the response.  If the response starts out with a copy of the request, it would incorrectly contain the password value.

Perhaps all of these items could be addressed by adding a responseHolder parameter to the Post function, which the caller must provide as a structure of the correct type without any values set.  What do you think of that idea?

It isn't clear to me how asynchronous responses (202 Accepted) will work, so maybe my suggestion won't be great for that case.  List actions could seemingly clone the responseHolder as necessary, though it isn't clear to me whether List is fully implemented yet.

vidhutsingh: > Subsequent code trusts that the "resp" value is an accessControlRecord structure. Why? Because the encodedPayload we pass in is an accessControlRecord structure, and Post returns the same type it is given. 

POST doesn't return the same type as it is given, it returns the interface{} type. Take a look at group_mgmt_client.go 

>By the same token, why wouldn't we trust that "resp" is not nil?

This is additional check to ensure it doesn't lead to application crash. 

>It isn't clear to me how asynchronous responses (202 Accepted) will work,

It is still not implemented. SDK is work in progress and there are many things in the pipeline to be implemented. 

>Perhaps all of these items could be addressed by adding a responseHolder parameter to the Post function, which the caller must provide as a structure of the correct type without any values set. What do you think of that idea?

I feel this is unnecessary as we don't need to pass the struct type .

vidhutsingh:
> How much effort would be required to put up a separate pull request with just the package rename changes, redone from scratch rather than as a modification from the current state of this pull request? If not so much effort, I'd suggest that as a first step. Then the remaining changes can be applied on top of that, and we can see exactly what they are. Also, if you agree with my speculation that separate "user facing" objects might not be needed, it should greatly simplify the remaining changes.

This review request was created a few weeks back and subsequent changes have been added to it due to delay on PR request. There are pending changes in my local repo which I have sent for review. I'm losing track of code changes. 

Next PR will have generated changes and specific code changes separately. But generated code changes will always be like this since it will have impact on all the objects.

vidhutsingh:
I am working on changeset which will remove this internal i.e EncodeAccessControlRecord/DcodeAccessControlRecord logic . Instead sdk user would have to  use conversion package which essentially converts the data to pointers.

pkg/client/access_control_records_object_set.go
rkumpf: > POST doesn't return the same type as it is given, it returns the interface{} type. Take a look at group_mgmt_client.go

I see that Post is given an interface{} and it returns an interface{}.  It doesn't document what they are or how they are related, but it looks to me like the returned type is the same as the given type.  How do they differ?

> > By the same token, why wouldn't we trust that "resp" is not nil?

> This is additional check to ensure it doesn't lead to application crash.

A failed type assertion also causes a panic, yes?

> I feel this is unnecessary as we don't need to pass the struct type .

I don't understand.  Is my suspicion correct that the returned structure might contain leftover attributes from the request?

vidhutsingh: You should check the POST implementation not simply the return type interface{}. POST is invoked by different objectset which has different types. Internally it uses DataWrapper to extract the POST response . Interface{} is the best way to handle in Go lang. Why would you want to invoke POST with a separate parameter as data type  along with the actual field?

rgcostea:
> Next PR will have generated changes and specific code changes separately. But generated code changes will always be like this since it will have impact on all the objects.

If that's the case, just file separate PRs for generator changes please.

pkg/client/access_control_records_object_set.go
rkumpf: I can see that I am not communicating well.  I apologize for that.

Notice how the Get function has an "intf" parameter even though it does not require an input structure.  The intf is essentially used as an empty structure of the correct type, into which the response is unmarshalled (using a DataWrapper, as you mentioned).  I'm suggesting that perhaps Post could have a parameter for the "full" request structure (as it does now) and a separate parameter for the "empty" response structure.  It would address my theorized concern about leftover request attributes being incorrectly included in the response structure.

Furthermore, I am wondering if it would make sense for the response structure to be modified in place, with the corresponding function return value removed.  The client could do something like this:

  requestPayload, err := nimbleos.EncodeAccessControlRecord(payload)
  ... (check err)
  responsePayload := &accessControlRecord{}
  err := objectSet.Client.Post(accessControlRecordPath, requestPayload, responsePayload)
  ... (check err)
  // Note: The DecodeAccessControlRecord parameter could be *accessControlRecord instead of interface{}.
  return nimbleos.DecodeAccessControlRecord(responsePayload)

There would be no way for responsePayload to be nil or to be of an incorrect type.  We wouldn't need logic to check for either, and we could remove the type assertion that could theoretically cause a panic.

vidhutsingh: >Notice how the Get function has an "intf" parameter even though it does not require an input structure. The intf is essentially used as an empty structure of the correct type, into which the response is unmarshalled (using a DataWrapper, as you mentioned). I'm suggesting that perhaps Post could have a parameter for the "full" request structure (as it does now) and a separate parameter for the "empty" response structure. It would address my theorized concern about leftover request attributes being incorrectly included in the response structure.

Now I see your point but that will only happen if API response is nil for the concerned attribute and HTTP return a SUCCESS since the same variable payload is used for unmarshalling as well. Although, it is a corner case but still valid. 

thanks for providing more details. I will fix it. We don't need to pass additional empty data argument to the POST function, we could simply re-initialized the payload variable in the POST implementation function that will address your concern. I will give a try with reflect otherwise we could pass this as a argument.

vidhutsingh:
Closing this pull request and will send separate PR as per the reviewer request.

vidhutsingh:
I am reopening the PR since there is no easy way now to create a separate PR for each changes in the current PR. 
Git doesn't allow you to create a new fork for existing user for the same upstream .

vidhutsingh:
@rkumpf , there is no easy way I could create a separate new pull request from the existing PR. I have reverted the changes of encoding/decoding . Also, I will address your review changes w.r.t POST/PUT in the new PR. I didn't want to club that change in this PR as it will add more complexity to this PR. Once this is approved, I will merge this change and from here on you should expect a minimal possible patch. 
If you need help, I can walk you through the changes.

pkg/client/v1/nimbleos/ns_n_i_c.go
```diff
@@ -1,18 +1,17 @@
...
-	SubnetLabelfield:= "subnet_label"
-	DataIpfield:= "data_ip"
-		
-	NsNICFields= &NsNIC{
+func init() {
+	Namefield := "name"
+	SubnetLabelfield := "subnet_label"
+	DataIpfield := "data_ip"
+
+	NsNICFields = &NsNIC{
```
rkumpf, line 14: When creating file names, I guess the generator adds an underscore before all capital letters.  It has some odd results like NsNic becoming ns_n_i_c when ns_nic would be better.  Other examples are ID and IDKV.  Perhaps we could suppress the underscore between consecutive capital letters if that does cause an incorrect result elsewhere.  Otherwise we could teach it to handle particular acronyms.

rkumpf (approved):
Thank you, this patch is much more straightforward to review.

pkg/client/v1/nimbleos/ns_n_i_c.go
vidhutsingh: Sure we can do that.

PR #6, opened 2020-07-24
Fixed POST/PUT object leftover attribute issue

vidhutsingh:
Bug 👍 SA-84
Problem : POST & PUT  function doesn't drop the object attribute if return nil by the Array .
Implementation 
Pass the pointer to concrete data type to POST & PUT function which would receive the response from the Array instead of using the same payload variable for request and response. 

Test:
--- PASS: TestVolumeServiceTestSuite (20.38s)
    --- PASS: TestVolumeServiceTestSuite/TestGetNonExistentVolumeByID (20.38s)
PASS
ok  	github.com/hpe-storage/nimble-golang-sdk/pkg/service	20.397s

### DIFF
--- a/pkg/client/access_control_records_object_set.go
+++ b/pkg/client/access_control_records_object_set.go
@@ -22,14 +22,9 @@ type AccessControlRecordObjectSet struct {
 
 // CreateObject creates a new AccessControlRecord object
 func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
-	resp, err := objectSet.Client.Post(accessControlRecordPath, payload)
+	resp, err := objectSet.Client.Post(accessControlRecordPath, payload, &nimbleos.AccessControlRecord{})
 	if err != nil {
 		return nil, err
-	}
-
-	// null check
-	if resp == nil {
-		return nil, nil
 	}
 
 	return resp.(*nimbleos.AccessControlRecord), err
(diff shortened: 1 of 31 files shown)


pkg/client/access_control_records_object_set.go
```diff
@@ -22,16 +22,11 @@ type AccessControlRecordObjectSet struct {
 
 // CreateObject creates a new AccessControlRecord object
 func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
-	resp, err := objectSet.Client.Post(accessControlRecordPath, payload)
+	resp, err := objectSet.Client.Post(accessControlRecordPath, payload, &nimbleos.AccessControlRecord{})
```
rkumpf, line 25: If we keep a handle to the response payload structure, we will be sure the Post result cannot cause a panic (if nil or incorrect type).  For example:

	responsePayload := &nimbleos.AccessControlRecord{}
	err := objectSet.Client.Post(accessControlRecordPath, payload, responsePayload)
	if err != nil {
		return nil, err
	}

	return responsePayload, err

pkg/client/group_mgmt_client.go
```diff
@@ -137,7 +137,7 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}) (interf
 		}
 		return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
 	}
-	return unwrap(response.Body(), payload)
+	return unwrap(response.Body(), respHolder)
 }
 
 // Get : Only used to get a single object with the given ID
```
rkumpf, line 143: Rename intf to respHolder for consistency?

rkumpf (approved):

pkg/client/group_mgmt_client.go
vidhutsingh: Sure, next PR.

PR #3, opened 2020-06-30
SDK consumable services implementations 

vidhutsingh:
Feature :
Implemented consumable services for all the NimOS endpoints 

Details :
1.Developed a template for generating all the NimOS services . The services contains generic method on objects e.g CREATE,MODIFY, DELETE and also contains methods by Name and by Id. It generated additional methods for Object type volume. 

2.Create a new template for generating group_service_factory.go which initialises the endpoint services . 

3.Change type from GroupService to NsGroupService as it was causing name conflict during Group object generation . 

4.Modified volume_service_test.go unit test module to accommodate above changes. 

Test :

XXXXXXXXXXXX$ go test
PASS
ok  	github.hpe.com/nimble-dcs/golang-sdk/pkg/service	42.530s

PR #2, opened 2020-06-24
Generated source code from template

vidhutsingh:
Feature : GoLangSDK 
Developed  templates for object, objectSet, enum, objectSetFactory for generation SDK client packages. 
1. object - 
	1.Handled complex data type, object , array of objects and array of 	complex datatypes .
 	2.Generated all the nimos objects.
2. objectSet -
    1. Generated all the objectSets method for the objects.  
   3. ObjectSetFactory 
	1. This contains the GetObjectSet methods (kind of constructor ) for 
	    all the objectSets i.e VolumeObjectSet(). 
   4. Enum 
	Creates a template for enum types and it generated all the nimos enum

Converted all the object’s member to pointer type as it was causing issue during communication to Array firmware. Our object contains both request and response members in the same object (structure) thus firmware was throwing error “SM_unexpected_arg” since it was not expecting response member during request. By using pointers types this is handled as nil pointers are dropped during API communication. 

Made changes to volume_service_test suite 

Test :
Go test . 
Able to perform all the basic operation on Array.

PR #7, opened 2020-07-28
Implemented pagination feature

vidhutsingh:
Feature: SDK pagination 

Implementation : 
Pagination feature allows users to process the objects in batch. Once the user sets the PageSize attribute,  the SDK would limit the no of return objects to PageSize. There is an option ContainsMore() to iterate over the complete list of objects. 
if no PageSize is specified, it would return all the objects from the Array. 

Test:
Implemented a unit test TestGetVolumesPagination() which validate the batch size request. 
    --- PASS: TestVolumeServiceTestSuite/TestGetNonExistentVolumeByID (10.57s)
    --- PASS: TestVolumeServiceTestSuite/TestGetVolumesPagination (10.24s)
PASS
ok  	github.com/hpe-storage/nimble-golang-sdk/pkg/service	20.821s

### DIFF
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -226,7 +226,6 @@ func (client *GroupMgmtClient) ListFromParams(path string, params *param.GetPara
 		return nil, err
 	}
 
-	// pagination, reset the start row and end row
 	if params != nil && params.Page != nil {
 		params.Page.TotalRows = wrapper.TotalRows
 	}
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -33,6 +33,7 @@ type DataWrapper struct {
 	StartRow      *int        `json:"startRow,omitempty"`
 	EndRow        *int        `json:"endRow,omitempty"`
 	PageSize      *int        `json:"pageSize,omitempty"`
+	TotalRows     *int        `json:"totalRows,omitempty"`
 	OperationType *string     `json:"operationType,omitempty"`
 }
 
@@ -219,14 +220,45 @@ func (client *GroupMgmtClient) List(path string) (interface{}, error) {
 
 // ListFromParams :
 func (client *GroupMgmtClient) ListFromParams(path string, params *param.GetParams) (interface{}, error) {
-	listResp, err := client.listGetOrPost(path, params)
+
+	// pagination, set the start row and end row if the pagination is enabled.
+	// TotalRows is nil if this is first pagination request.
+	if params != nil && params.Page != nil {
+		// TotalRows is nil for first request
+		if params.Page.TotalRows == nil {
+			//check if page size is set
+			if params.Page.PageSize != nil {
+				if params.Page.StartRow == nil {
+					params.Page.SetStartRow(0)
+				}
+				params.Page.SetEndRow(*params.Page.PageSize)
+			}
+		} else {
+			params.Page.SetStartRow(*params.Page.EndRow)
+
+			totalRows := *params.Page.StartRow + *params.Page.PageSize
+			// overflow check
+			if totalRows > *params.Page.TotalRows {
+				totalRows = *params.Page.TotalRows
+			}
+			params.Page.SetEndRow(totalRows)
+
+		}
+	}
+
+	wrapper, err := client.listGetOrPost(path, params)
 	if err != nil {
 		return nil, err
 	}
-	return listResp, nil
+
+	// pagination, reset the start row and end row
+	if params != nil && params.Page != nil {
+		params.Page.TotalRows = wrapper.TotalRows
+	}
+	return wrapper.Data, nil
 }
 
-func (client *GroupMgmtClient) listGetOrPost(path string, params *param.GetParams) (interface{}, error) {
+func (client *GroupMgmtClient) listGetOrPost(path string, params *param.GetParams) (*DataWrapper, error) {
 	if params == nil {
 		return client.listGet(path, nil)
 	}
@@ -244,9 +276,8 @@ func (client *GroupMgmtClient) listGetOrPost(path string, params *param.GetParam
 			fetch := "fetch"
 			wrapper := &DataWrapper{
 				Data:          params.Filter,
-				StartRow:      params.StartRow,
-				EndRow:        params.EndRow,
-				PageSize:      params.PageSize,
+				StartRow:      params.Page.StartRow,
+				EndRow:        params.Page.EndRow,
 				OperationType: &fetch,
 			}
 			// complex filter, need to POST it
@@ -278,7 +309,7 @@ func (client *GroupMgmtClient) listPost(
 	path string,
 	payload *DataWrapper,
 	queryParams map[string]string,
-) (interface{}, error) {
+) (*DataWrapper, error) {
 	// build the url
 	url := fmt.Sprintf("%s/%s/detail", client.URL, path)
 
@@ -310,14 +341,14 @@ func (client *GroupMgmtClient) listPost(
 		return nil, err
 	}
 	// return it
-	return wrapper.Data, nil
+	return wrapper, nil
 }
 
 // listGet uses a get request to get all objects on the path
 func (client *GroupMgmtClient) listGet(
 	path string,
 	queryParams map[string]string,
-) (interface{}, error) {
+) (*DataWrapper, error) {
 	// build the url
 	url := fmt.Sprintf("%s/%s/detail", client.URL, path)
 
@@ -346,7 +377,7 @@ func (client *GroupMgmtClient) listGet(
 	}
 
 	// return it
-	return wrapper.Data, nil
+	return wrapper, nil
 }
 
 // unwrap a response body
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -221,31 +221,6 @@ func (client *GroupMgmtClient) List(path string) (interface{}, error) {
 // ListFromParams :
 func (client *GroupMgmtClient) ListFromParams(path string, params *param.GetParams) (interface{}, error) {
 
-	// pagination, set the start row and end row if the pagination is enabled.
-	// TotalRows is nil if this is first pagination request.
-	if params != nil && params.Page != nil {
-		// TotalRows is nil for first request
-		if params.Page.TotalRows == nil {
-			//check if page size is set
-			if params.Page.PageSize != nil {
-				if params.Page.StartRow == nil {
-					params.Page.SetStartRow(0)
-				}
-				params.Page.SetEndRow(*params.Page.PageSize)
-			}
-		} else {
-			params.Page.SetStartRow(*params.Page.EndRow)
-
-			totalRows := *params.Page.StartRow + *params.Page.PageSize
-			// overflow check
-			if totalRows > *params.Page.TotalRows {
-				totalRows = *params.Page.TotalRows
-			}
-			params.Page.SetEndRow(totalRows)
-
-		}
-	}
-
 	wrapper, err := client.listGetOrPost(path, params)
 	if err != nil {
 		return nil, err
--- a/pkg/param/get_params.go
+++ b/pkg/param/get_params.go
@@ -43,9 +43,6 @@ func (params *GetParams) BuildQueryParts() (map[string]string, error) {
 		}
 		queryParts["fields"] = strings.Join(params.Fields, ",")
 	}
-	//if params.PageSize != nil {
-	//queryParts["pageSize"] = fmt.Sprintf("%d", *params.PageSize)
-	//}
 
 	if params.SortBy != nil {
 		var sortByStrings []string
--- a/pkg/param/get_params.go
+++ b/pkg/param/get_params.go
@@ -4,16 +4,15 @@ package param
 
 import (
 	"fmt"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param/pagination"
 	"strings"
 )
 
 type GetParams struct {
-	Filter   *SearchFilter
-	Fields   []string
-	StartRow *int
-	EndRow   *int
-	PageSize *int
-	SortBy   []SortOrder
+	Filter *SearchFilter
+	Fields []string
+	Page   *pagination.GetPage
+	SortBy []SortOrder
 }
 
 func (params *GetParams) WithFields(fields []string) {
@@ -30,12 +29,12 @@ func (params *GetParams) WithSortBy(sortBy []SortOrder) {
 
 func (params *GetParams) BuildQueryParts() (map[string]string, error) {
 	queryParts := make(map[string]string)
-	if params.StartRow != nil {
-		queryParts["startRow"] = fmt.Sprintf("%d", *params.StartRow)
+	if params.Page != nil && params.Page.StartRow != nil {
+		queryParts["startRow"] = fmt.Sprintf("%d", *params.Page.StartRow)
 	}
 
-	if params.EndRow != nil {
-		queryParts["endRow"] = fmt.Sprintf("%d", *params.EndRow)
+	if params.Page != nil && params.Page.EndRow != nil {
+		queryParts["endRow"] = fmt.Sprintf("%d", *params.Page.EndRow)
 	}
 
 	if params.Fields != nil {
@@ -44,9 +43,9 @@ func (params *GetParams) BuildQueryParts() (map[string]string, error) {
 		}
 		queryParts["fields"] = strings.Join(params.Fields, ",")
 	}
-	if params.PageSize != nil {
-		queryParts["pageSize"] = fmt.Sprintf("%d", *params.PageSize)
-	}
+	//if params.PageSize != nil {
+	//queryParts["pageSize"] = fmt.Sprintf("%d", *params.PageSize)
+	//}
 
 	if params.SortBy != nil {
 		var sortByStrings []string
--- a/pkg/param/get_params.go
+++ b/pkg/param/get_params.go
@@ -11,7 +11,7 @@ import (
 type GetParams struct {
 	Filter *SearchFilter
 	Fields []string
-	Page   *pagination.GetPage
+	Page   *pagination.Page
 	SortBy []SortOrder
 }
 
--- a/pkg/param/pagination/pagination.go
+++ b/pkg/param/pagination/pagination.go
@@ -1,5 +1,6 @@
 package pagination
 
+// GetPage - pagination
 type GetPage struct {
 	StartRow  *int
 	EndRow    *int
@@ -7,6 +8,7 @@ type GetPage struct {
 	TotalRows *int
 }
 
+// ContainsMore - returns true if there are more pending objects then page size.
 func (page *GetPage) ContainsMore() bool {
 
 	if *page.EndRow < *page.TotalRows {
@@ -14,17 +16,18 @@ func (page *GetPage) ContainsMore() bool {
 	}
 	return false
 }
+
+// SetStartRow - set the starting row from list of objects
 func (page *GetPage) SetStartRow(row int) {
 	page.StartRow = &row
 }
 
+// SetEndRow - set the end row from the list of the return objects
 func (page *GetPage) SetEndRow(row int) {
 	page.EndRow = &row
 }
-func (page *GetPage) SetTotalRows(row int) {
-	page.TotalRows = &row
-}
 
+// SetPageSize - set the page size for batch processing
 func (page *GetPage) SetPageSize(row int) {
 	page.PageSize = &row
 }
--- a/pkg/param/pagination/pagination.go
+++ b/pkg/param/pagination/pagination.go
@@ -1,7 +1,9 @@
 package pagination
 
-// GetPage - pagination
-type GetPage struct {
+// Page - This is used for batch processing. User may want to process
+// limited objects instead of processing the whole object list . Specially when there are
+// 100s of objects exists on Array e.g. volumes, snapshots etc.
+type Page struct {
 	StartRow  *int
 	EndRow    *int
 	PageSize  *int
@@ -9,25 +11,22 @@ type GetPage struct {
 }
 
 // ContainsMore - returns true if there are more pending objects then page size.
-func (page *GetPage) ContainsMore() bool {
+func (page *Page) ContainsMore() bool {
+	return *page.EndRow < *page.TotalRows
 
-	if *page.EndRow < *page.TotalRows {
-		return true
-	}
-	return false
 }
 
 // SetStartRow - set the starting row from list of objects
-func (page *GetPage) SetStartRow(row int) {
+func (page *Page) SetStartRow(row int) {
 	page.StartRow = &row
 }
 
 // SetEndRow - set the end row from the list of the return objects
-func (page *GetPage) SetEndRow(row int) {
+func (page *Page) SetEndRow(row int) {
 	page.EndRow = &row
 }
 
 // SetPageSize - set the page size for batch processing
-func (page *GetPage) SetPageSize(row int) {
+func (page *Page) SetPageSize(row int) {
 	page.PageSize = &row
 }
--- a/pkg/param/pagination/pagination.go
+++ b/pkg/param/pagination/pagination.go
@@ -1,32 +1,46 @@
 package pagination
 
-// Page - This is used for batch processing. User may want to process
-// limited objects instead of processing the whole object list . Specially when there are
-// 100s of objects exists on Array e.g. volumes, snapshots etc.
+// Page controls enable iterative retrieval of a large data set. This
+// structure is used both on input, to specify the boundaries of a
+// page to retrieve, and on output, to indicate the items actually
+// retrieved. The TotalRows value is ignored on input
 type Page struct {
 	StartRow  *int
 	EndRow    *int
-	PageSize  *int
 	TotalRows *int
 }
 
-// ContainsMore - returns true if there are more pending objects then page size.
-func (page *Page) ContainsMore() bool {
-	return *page.EndRow < *page.TotalRows
-
-}
-
-// SetStartRow - set the starting row from list of objects
+// SetStartRow - Specify the first row to retrieve in a page of results, using a zero-based index.
 func (page *Page) SetStartRow(row int) {
 	page.StartRow = &row
 }
 
-// SetEndRow - set the end row from the list of the return objects
+// SetEndRow - Specify the first row NOT to retrieve in a page of results, using a zero-based index
+// value one larger than the last row desired.
 func (page *Page) SetEndRow(row int) {
 	page.EndRow = &row
 }
 
 // SetPageSize - set the page size for batch processing
-func (page *Page) SetPageSize(row int) {
-	page.PageSize = &row
+func (page *Page) SetPageSize(psize int) {
+	zero := 0
+	page.StartRow = &zero
+	page.EndRow = &psize
+}
+
+// NextPage - Returns true if objects are more than pagesize.
+func (page *Page) NextPage() bool {
+	if page.StartRow == nil || page.EndRow == nil || page.TotalRows == nil ||
+		*page.EndRow >= *page.TotalRows || *page.StartRow >= *page.EndRow {
+		return false
+	}
+	pageSize := *page.EndRow - *page.StartRow
+	page.StartRow = page.EndRow
+	// overflow check
+	if (*page.StartRow + pageSize) > *page.TotalRows {
+		page.SetEndRow(*page.TotalRows)
+	} else {
+		page.SetEndRow(*page.StartRow + pageSize)
+	}
+	return true
 }
--- a/pkg/param/pagination/pagination.go
+++ b/pkg/param/pagination/pagination.go
@@ -2,7 +2,7 @@ package pagination
 
 // Page controls enable iterative retrieval of a large data set. This
 // structure is used both on input, to specify the boundaries of a
-// page to retrieve, and on output, to indicate the items actually
+// page to retrieve, and on output, to indicate the no of items actually
 // retrieved. The TotalRows value is ignored on input
 type Page struct {
 	StartRow  *int
@@ -21,14 +21,16 @@ func (page *Page) SetEndRow(row int) {
 	page.EndRow = &row
 }
 
-// SetPageSize - set the page size for batch processing
+// SetPageSize - Sets the StartRow and EndRow values to request the first page of the specified size.
 func (page *Page) SetPageSize(psize int) {
 	zero := 0
 	page.StartRow = &zero
 	page.EndRow = &psize
 }
 
-// NextPage - Returns true if objects are more than pagesize.
+// NextPage - Updates the StartRow and EndRow values to request the next page of the
+// same size, if more rows are available according to the TotalRows value.
+// Returns true if the values are actually updated.
 func (page *Page) NextPage() bool {
 	if page.StartRow == nil || page.EndRow == nil || page.TotalRows == nil ||
 		*page.EndRow >= *page.TotalRows || *page.StartRow >= *page.EndRow {
--- a/pkg/service/volume_service_test.go
+++ b/pkg/service/volume_service_test.go
@@ -110,25 +110,17 @@ func (suite *VolumeServiceTestSuite) TestGetVolumesPagination() {
 
 	arg.Page = pagination
 	// fetch only 2 volumes
-	volumes, err := suite.volumeService.GetVolumes(arg)
-	if err != nil {
-		suite.T().Errorf("TestGetVolumesPagination(): Unable to fetch volumes, err: %v", err.Error())
-		return
-	}
-
-	if len(volumes) != 2 {
-		suite.T().Errorf("TestGetVolumesPagination(): pagination count doesnt match with # of return volumes ")
-		return
-	}
-
-	// Batch processing
-	for arg.Page.NextPage() {
+	for hasNextPage := true; hasNextPage; hasNextPage = arg.Page.NextPage() {
 		volumes, err := suite.volumeService.GetVolumes(arg)
 		if err != nil {
 			suite.T().Errorf("TestGetVolumesPagination(): Unable to fetch volumes, err: %v", err.Error())
 			return
 		}
-		_= volumes
+
+		if len(volumes) > 2 {
+			suite.T().Errorf("TestGetVolumesPagination(): Returned volumes are more than the requested page size")
+			return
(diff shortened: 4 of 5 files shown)


pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
 package pagination
 
+// GetPage - pagination
 type GetPage struct {
 	StartRow  *int
 	EndRow    *int
 	PageSize  *int
```
rkumpf: Why is a PageSize included in the model?  It doesn't appear to be included in the Nimble REST API, and it seems redundant with StartRow/EndRow.  Could we remove it?

I get the idea we are trying to present a simpler interface to SDK users than is provided by the Nimble REST API.  If this is a goal, we'll want to be thoughtful about how we do it.  Perhaps we can provide convenience functions that modify the values of the StartRow and EndRow fields without using additional fields.

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
...
 	TotalRows *int
 }
 
+// ContainsMore - returns true if there are more pending objects then page size.
 func (page *GetPage) ContainsMore() bool {
 
 	if *page.EndRow < *page.TotalRows {
```
rkumpf: Simplify:

  return *page.EndRow < *page.TotalRows

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
...
+// GetPage - pagination
 type GetPage struct {
 	StartRow  *int
 	EndRow    *int
 	PageSize  *int
 	TotalRows *int
 }
 
+// ContainsMore - returns true if there are more pending objects then page size.
 func (page *GetPage) ContainsMore() bool {
```
rkumpf: It seems clearer to name it isLastPage() (reversing the sense) or hasNextPage().

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
 package pagination
 
+// GetPage - pagination
 type GetPage struct {
```
rkumpf: We seem to be using the same structure for input and output, so "Get" doesn't seem right.  Should we just call it "Page"?

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
...
+// ContainsMore - returns true if there are more pending objects then page size.
 func (page *GetPage) ContainsMore() bool {
 
 	if *page.EndRow < *page.TotalRows {
 		return true
 	}
 	return false
 }
+
+// SetStartRow - set the starting row from list of objects
```
rkumpf: "from list of objects" implies this field is only set for output.  That doesn't seem like what we'd want.

pkg/param/get_params.go
```diff
@@ -44,9 +43,9 @@ func (params *GetParams) BuildQueryParts() (map[string]string, error) {
 		}
 		queryParts["fields"] = strings.Join(params.Fields, ",")
 	}
-	if params.PageSize != nil {
-		queryParts["pageSize"] = fmt.Sprintf("%d", *params.PageSize)
-	}
+	//if params.PageSize != nil {
```
rkumpf: Remove dead code?

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
 package pagination
 
+// GetPage - pagination
 type GetPage struct {
 	StartRow  *int
```
rkumpf, line 8: Would it be better to use uint?

pkg/client/group_mgmt_client.go
```diff
@@ -219,14 +220,45 @@ func (client *GroupMgmtClient) List(path string) (interface{}, error) {
...
+				if params.Page.StartRow == nil {
+					params.Page.SetStartRow(0)
+				}
+				params.Page.SetEndRow(*params.Page.PageSize)
+			}
+		} else {
+			params.Page.SetStartRow(*params.Page.EndRow)
+
+			totalRows := *params.Page.StartRow + *params.Page.PageSize
+			// overflow check
```
rkumpf: I don't feel this check is needed.  The client's job is to ask for what they want, and the server's job is to return what it has.  If the client asks for more than exist, the server should only return what exists.  Consider the first page case.  At line 234, we might set the EndRow value too high.  It shouldn't matter.

pkg/client/group_mgmt_client.go
```diff
@@ -219,14 +220,45 @@ func (client *GroupMgmtClient) List(path string) (interface{}, error) {
 
 // ListFromParams :
 func (client *GroupMgmtClient) ListFromParams(path string, params *param.GetParams) (interface{}, error) {
-	listResp, err := client.listGetOrPost(path, params)
+
+	// pagination, set the start row and end row if the pagination is enabled.
+	// TotalRows is nil if this is first pagination request.
```
rkumpf: I'm not sure the "move to next page" logic belongs here.  The caller to this function should provide the parameters for what it wants, not what was returned from the previous call.

Maybe what we need is a function like this:

func (page *GetPage) GetNextPage() GetPage {
  return Page{
    StartRow: page.EndRow
    EndRow: page.EndRow - pageStartRow + pageEndRow
  }
}

or this:

func (page *GetPage) Advance() {
  pageSize := page.EndRow - page.StartRow
  page.StartRow = page.EndRow
  page.EndRow = page.EndRow + pageSize
}

pkg/client/group_mgmt_client.go
```diff
@@ -219,14 +220,45 @@ func (client *GroupMgmtClient) List(path string) (interface{}, error) {
...
+		}
+	}
+
+	wrapper, err := client.listGetOrPost(path, params)
 	if err != nil {
 		return nil, err
 	}
-	return listResp, nil
+
+	// pagination, reset the start row and end row
```
rkumpf: We should copy StartRow and EndRow from the wrapper back to the params, right?

I'm not sure about the Nimble API, but in some APIs the server is not obligated to return as many rows as the client asks for, if it thinks the request is excessive.

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,33 @@
 package pagination
 
+// GetPage - pagination
```
rkumpf: Let's add clarifying comments regarding how this structure is used in requests and in responses.

pkg/service/volume_service_test.go
```diff
@@ -26,7 +27,7 @@ func (suite *VolumeServiceTestSuite) SetupTest() {
 	}
 
 	// set debug
-	//groupService.SetDebug()
+	groupService.SetDebug()
```
rkumpf: Should this be submitted?

pkg/service/volume_service_test.go
```diff
@@ -101,6 +102,50 @@ func (suite *VolumeServiceTestSuite) TestGetNonExistentVolumeByID() {
...
+	for arg.Page.ContainsMore() {
+		volumes, err := suite.volumeService.GetVolumes(arg)
+		if err != nil {
+			suite.T().Errorf("TestGetVolumesPagination(): Unable to fetch volumes, err: %v", err.Error())
+			return
+		}
+		fmt.Printf("contains more volumes , parmas %+v, volumes : %v", arg, volumes)
+	}
+
+	// dont set batch size, should return all the volumes
```
rkumpf, line 126: Should it be a separate test case/function?

pkg/service/volume_service_test.go
```diff
@@ -101,6 +102,50 @@ func (suite *VolumeServiceTestSuite) TestGetNonExistentVolumeByID() {
...
+	}
+
+	// Batch processing
+	for arg.Page.ContainsMore() {
+		volumes, err := suite.volumeService.GetVolumes(arg)
+		if err != nil {
+			suite.T().Errorf("TestGetVolumesPagination(): Unable to fetch volumes, err: %v", err.Error())
+			return
+		}
+		fmt.Printf("contains more volumes , parmas %+v, volumes : %v", arg, volumes)
```
rkumpf: Typo *params

rkumpf (changes requested):

pkg/param/pagination/pagination.go
vidhutsingh: That already exist, you can set StartRow and EndRow .

pkg/param/pagination/pagination.go
vidhutsingh: Done.

pkg/param/pagination/pagination.go
vidhutsingh: isLastPage() is not appropriate here , we can do hasNextPage() but 
this is in line with other sdk implementation where we have such fn containerMore().

pkg/param/pagination/pagination.go
vidhutsingh: Yes, thanks .

pkg/param/pagination/pagination.go
vidhutsingh: This is used to set the starting row for the object list and it is meant for sdk user

pkg/client/group_mgmt_client.go
vidhutsingh: Yes if there are fewer objects it is not obliged and that he handled as well. pls check line 240

pkg/param/pagination/pagination.go
vidhutsingh: Sure .

pkg/service/volume_service_test.go
vidhutsingh: We can omit it .

pkg/service/volume_service_test.go
vidhutsingh: There is no strict rule since this is part of pagination so I kept it in a single function.

pkg/service/volume_service_test.go
vidhutsingh: Sure . thanks

pkg/param/pagination/pagination.go
```diff
@@ -1,33 +1,32 @@
 package pagination
 
-// GetPage - pagination
-type GetPage struct {
+// Page - This is used for batch processing. User may want to process
```
rkumpf: I suggest:

// Page controls enable iterative retrieval of a large data set.  This
// structure is used both on input, to specify the boundaries of a
// page to retrieve, and on output, to indicate the items actually
// retrieved.  The TotalRows value is ignored on input.

pkg/param/pagination/pagination.go
rkumpf: I suggest:

// Specify the first row to retrieve in a page of results, using a zero-based index.

pkg/param/pagination/pagination.go
```diff
@@ -1,33 +1,32 @@
...
 }
 
 // SetStartRow - set the starting row from list of objects
-func (page *GetPage) SetStartRow(row int) {
+func (page *Page) SetStartRow(row int) {
 	page.StartRow = &row
 }
 
 // SetEndRow - set the end row from the list of the return objects
-func (page *GetPage) SetEndRow(row int) {
```
rkumpf, line 21: I suggest:

// Specify the first row NOT to retrieve in a page of results, using a zero-based index
// value one larger than the last row desired.

pkg/param/pagination/pagination.go
rkumpf: Why do we want multiple ways to specify the request parameters?  The meaning isn't clear if all the values are specified.

From the test case, the expected flow appears to be:

    arg.Page = new(pagination.Page)
    arg.Page.SetPageSize(2)
    volumes, err := suite.volumeService.GetVolumes(arg)
    for arg.Page.ContainsMore() {
            volumes, err := suite.volumeService.GetVolumes(arg)
    }

I find it unexpected that the GetVolumes function modifies the arg value.  We could clarify it by explicitly returning the Page values from the response.  I suggest a flow like:

	arg.Page = new(pagination.Page)
	var retrievedPage pagination.Page
	for hasNextPage := true; hasNextPage; hasNextPage = retrievedPage.HasNextPage() {
		arg.Page.NextPage(retrievedPage, 2)
		volumes, retrievedPage, err := suite.volumeService.GetVolumes(arg)
	}

The supporting function could look something like:

    func (page *Page) NextPage(retrievedPage Page, pageSize int) {
    	if retrievedPage == nil || retrievedPage.EndRow == nil {
    		page.SetStartRow(0)
    	} else {
    		page.SetStartRow(*retrievedPage.EndRow)
    	}
    	page.SetEndRow(*page.StartRow + pageSize)
    }

pkg/param/pagination/pagination.go
```diff
@@ -1,33 +1,32 @@
...
+type Page struct {
 	StartRow  *int
 	EndRow    *int
 	PageSize  *int
 	TotalRows *int
 }
 
 // ContainsMore - returns true if there are more pending objects then page size.
-func (page *GetPage) ContainsMore() bool {
+func (page *Page) ContainsMore() bool {
```
rkumpf: With the flow I suggested in an earlier comment, I'm not sure we gain much by overloading the Page structure for use in both input and output.  We could use separate structures, with the TotalRows attribute and this function as part of just the output structure.  It would align better with the model of the underlying Nimble REST API.  I'm not sure either way is significantly better than the other, but it seems worth considering.

pkg/service/volume_service_test.go
rkumpf: Okay.  Generally I believe it is considered good practice to keep unit test cases small and targeted.  GoLang seems to prefer test cases defined in a structured list of inputs and outputs, though that might not be straightforward to apply here.

pkg/param/pagination/pagination.go
rkumpf: Where can I find the precedent you refer to?

pkg/param/pagination/pagination.go
vidhutsingh: >>Why do we want multiple ways to specify the request parameters? The meaning isn't clear if all the values are specified.
We are exposing only methods to set the page size , start row and end row for end user .  We arent exposing convenience method for TotalRow as it is not expected to be set by user . 

StartRow and EndRow add more flexibility to sdk user that's all. If we don't want we can remove it . 

`arg.Page = new(pagination.Page)
var retrievedPage pagination.Page
for hasNextPage := true; hasNextPage; hasNextPage = retrievedPage.HasNextPage() {
	arg.Page.NextPage(retrievedPage, 2)
	volumes, retrievedPage, err := suite.volumeService.GetVolumes(arg)
}`

This code will not work. As retrievedPage is a new variable. HasNextPage() will be false all the time. 

What is the use of this arg.Page.NextPage(retrievedPage, 2) , this is unnecessary . Intentions of having ContainsMore() is to jsut fetch the next Page so that user easily iterative over it.

pkg/param/pagination/pagination.go
vidhutsingh: Python and Perl SDK both have it . here is reference build/rest-sdk/sdk/python/cfl_autogen_rest_classes/borg_cmd_driver_rest_base/shelf_base.py:            while target_spec.getPaginationSpec().containsMore(): 

You can check SM repo

pkg/param/pagination/pagination.go
vidhutsingh: ok

pkg/param/pagination/pagination.go
vidhutsingh: Sure.

pkg/param/pagination/pagination.go
vidhutsingh: ok, thanks

pkg/param/pagination/pagination.go
rkumpf: I should have been more clear.  In group_mgmt_client.go, the end of the ListFromParams function would look something like:

	returnedPage := pagination.Page { &wrapper.StartRow, &wrapper.EndRow, &wrapper.TotalRows }
	return wrapper.Data, returnedPage, nil

It is a question of who has control over which page is retrieved.  I feel the "advance to next page" logic should not be hidden inside the ListFromParams function.  Rather, the caller should explicitly indicate the desired page parameters.  The Page.NextPage function provides a convenience to set the StartRow and EndRow based on the previously retrieved page and the desired page size.  Note that the suggested client code is just as compact, but it gives the client more control and improved clarity.

pkg/param/pagination/pagination.go
rkumpf: If there is relevant precedent, that's fine.  It's just a name.  I'm not sure where to find the "SM repo", though.  Could you point me to it, please?  Is it open source?  I guess you aren't referring to the Python SDK at https://github.com/hpe-storage/nimble-python-sdk.

pkg/param/pagination/pagination.go
vidhutsingh: We dont want to give the user any selective page control as this may not be useful and he may not need it. TotalRow is internal to SDK as it is based on REST API return. 
When you make a call to REST API, it returns TotalRows . 
Let's consider an example: let's say there are 100s of volume on the array and the user wants to process all of them so instead of fetching all the volumes at one time, the user would set pageSize = 10, he would iterate over 10 volumes at a time and process. 
Tell me the use case where the user would like to retrieve a selective page? 

Here pagination is not used for searching any selective page so why bother to give this control to the user.

pkg/param/pagination/pagination.go
vidhutsingh: It should be there in python sdk as well.

pkg/param/pagination/pagination.go
rkumpf: Is it the goal for this SDK to facilitate access to the underlying Nimble REST API, or to abstract it for particular use cases?

vidhutsingh: Pagination is a SDK feature.

rgcostea: @rkumpf , we want the SDK to make it easy to access the underlying storage resources.  Ideally, if the user simply requests all objects, the SDK ought to internally page and report all objects.  If the user specifies paging details, we should apply that to our API request and report paging details in the response.

rkumpf: Thanks, @rgcostea.  Interesting point about internally looping to fulfill the request page parameters if the array is unwilling to do so in a single request.  It does not appear to be part of the current design.  The same principle could apply when a caller requests rows 10 to 10,000,000, for example.  Perhaps a caller could explicitly request this convenience, to indicate acceptance of the response time implications.

This discussion concerns the page-level interface exposed by the SDK.  My aim is to avoid what I perceive as confusing redundancy in the page parameters while enabling a user to access the capabilities of the underlying API.  I.e., make it simple to loop through all the pages while still providing control to request rows 10 to 30, for example.

To achieve this, I'm suggesting: 1) Separate the retrieved page info from the requested page parameters (instead of modifying the input parameters with output data), 2) Move the PageSize attribute to a convenience function which can translate it to StartRow and EndRow, and 3) Give the caller control over subsequent pages rather than making assumptions inside the SDK.

rgcostea: > It does not appear to be part of the current design

It needs to be in the final product.  We do this in our python and java SDKs.

I feel all 3 of your suggestions are sound.  Keeping the SDK as simple as possible for the end user should be one of our overarching goals.

vidhutsingh: How could implementing this be a simplified approach :
```
arg.Page = new(pagination.Page)
var retrievedPage pagination.Page

for hasNextPage := true; hasNextPage; hasNextPage = retrievedPage.HasNextPage() {
	arg.Page.NextPage(retrievedPage, 2)
	volumes, retrievedPage, err := suite.volumeService.GetVolumes(arg)
}
```
User has to constantly feed the page no which is un-necessary, are you suggesting to have a int variable and increment it inside the for loop to set the page no all the time ? This is not at all intuitive to user.  

Here is an example of C++ SDK and ours implementation similar to that 
```// "name"=="0" && "size">=1024  sortBy:size
    gp = GetParamsSmtPtr(new GetParams(sf3, sk), getParamsDeleter);
    PaginationParamsSmtPtr pgp = gp->getPaginationParams();
    pgp->setBatchSize(2);
    pgp->setPaginationStartIndex(3);
    gp->setPaginationParams(pgp);
    do {
        volList = volumeOS->getObjectList(gp);
        startRow = pgp->getPaginationStartIndex();
        endRow = pgp->getPaginationEndIndex();
        totalRows = pgp->getTotalRowCount();
        cout << "startRow: " << startRow << "   endRow: " << endRow;
        cout << "   totalRows: " << totalRows << "   batchSize: " << pgp->getBatchSize() << endl;
        //ensure that the number of rows returned never exceeds the specfied batchSize.
        assert((pgp->getPaginationEndIndex() - pgp->getPaginationStartIndex()) <= pgp->getBatchSize());
    } while(pgp->containsMore());`
```
I could have written the same unit test code like this with existing pagination package 
```
gp := new(param.GetParams)

pgp :=new(pagination.Page)
pgp.SetPageSize(2)

gp.Page = pgp
volumes, err := suite.volumeService.GetVolumes(gp)
for  pgp.ContainsMore() {
	volumes, err := suite.volumeService.GetVolumes(gp)
}
```

rkumpf: The integer literal in the flow I suggested is a page size, not a page number.  A caller is not expected to track page numbers.

My principal aim is to clarify the interface.  For example, it is not obvious that this code would fetch rows 5 to 14:

    gp := new(param.GetParams)
    pgp :=new(pagination.Page)
    pgp.SetStartRow(5)
    pgp.SetEndRow(12)
    pgp.SetPageSize(10)
    gp.Page = pgp
    volumes, err := suite.volumeService.GetVolumes(gp)

And who would guess that this code would fetch rows 12 to 19?

    gp := new(param.GetParams)
    pgp :=new(pagination.Page)
    pgp.SetStartRow(5)
    pgp.SetEndRow(12)
    pgp.SetPageSize(10)
    x := 20
    pgp.totalRows = &x
    gp.Page = pgp
    volumes, err := suite.volumeService.GetVolumes(gp)

This one will panic:

    gp := new(param.GetParams)
    pgp :=new(pagination.Page)
    x := 20
    pgp.totalRows = &x
    gp.Page = pgp
    volumes, err := suite.volumeService.GetVolumes(gp)

One could argue "don't do that".  But in a general purpose SDK, and in open source no less, it is expected that the behavior is well defined.  Rather than document potentially bizarre behavior or limitations, it seems straightforward to remove the uncertainty from the interface.

I suggested a possible design to address the clarity concerns.  I am certainly open to alternatives.

vidhutsingh: The user input validation logic is not built yet since this is something that should be done for all the interfacing functions /object. so kept it open for now . That will clarify all of these to sdk user .   
Here in case of pagination, either you set PageSize and StartRow, or you set only PageSize or you set StartRow and EndRow that user input validation will be implemented before end product release. Currently it is not done , it may be causing confusion. 
The rest of the combination will throw error like PageSize , StartRow and EndRow. Only the above three listed combinations are allowed. 
So ,
```
gp := new(param.GetParams)
pgp :=new(pagination.Page)
x := 20
pgp.totalRows = &x
gp.Page = pgp
volumes, err := suite.volumeService.GetVolumes(gp)
```
Once user input validation is implemented, this will be gracefully handled and user would be notified with error message . 
If you notice I have provided only user settable options for PageSize , Start Row and Endrow . The TotalRow is not even allowed to be set by convenience function.

rkumpf: This caller code construction was posited:

	for  pgp.ContainsMore() {
		volumes, err := suite.volumeService.GetVolumes(gp)
	}

I feel the interface does not make it clear that the pgp values are modified by the GetVolumes() call, both before they are used in a REST API call and after a response is received.  The conflation of input parameters and output results seems unexpected and suboptimal.

vidhutsingh: @rkumpf, as I see you have only issue with the conflation of input/output parameters. As I posted earlier, this is similar to what we do in CPP SDK and maybe in others. I posted the example as well. How do you propose to change it ?

rkumpf: I expressed these concerns:

* Redundant/overlapping attributes in the Page model.
* Input parameters modified with "next page" logic that is hidden from the caller and outside their control.
* Input parameters unexpectedly modified with output values.

My suggestions to address these concerns are:

1. In ListFromParams (and in the Get functions that call it), add a third return value with the retrieved page values.  Do not modify the input parameters.
1. Remove the PageSize attribute from the Page structure.
1. Add a convenience function that sets appropriate StartRow and EndRow values in a Page structure based on the page values from a prior Get call (if applicable) and a given page size.  The NextPage function in an earlier comment is a possible implementation.
1. In ListFromParams, use the StartRow and EndRow values provided as input, without modification.

vidhutsingh: #1 )I am not convinced with your suggestion, earlier you said there is a conflation of input/output and as per your suggestion, we are endup doing the same thing in a different way. Here is your example
```
func (page *Page) NextPage(retrievedPage Page, pageSize int) {
	if retrievedPage == nil || retrievedPage.EndRow == nil {
		page.SetStartRow(0)
	} else {
		page.SetStartRow(*retrievedPage.EndRow)
	}
	page.SetEndRow(*page.StartRow + pageSize)
}
```
Aren't we modifying the in StartRow and EndRow here internally? we are modifying the input and output here . So conflation of i/o remains. 

#2 , The suggested change will have an additional return type retrivedPage
```
volumes, retrievedPage, err := suite.volumeService.GetVolumes(arg)
```

The paginartion is optional feature and user may not wish to set this , so this third argument is always going to be redundant and will have to pass nil all the time un-necessarily. In addition to it , this will have impact on all the services /object and objectset . I don't see much value here .

#3)

User has to make additional call to the fetch the nextPage() , here in your code example 
```
arg.Page = new(pagination.Page)
var retrievedPage pagination.Page
for hasNextPage := true; hasNextPage; hasNextPage = retrievedPage.HasNextPage() {
	arg.Page.NextPage(retrievedPage, 2)
	volumes, retrievedPage, err := suite.volumeService.GetVolumes(arg)
}
```
I don't think this is required. The approach I implemented is neat and easily consumable.
Here is an example 
```
arg.Page = new(pagination.Page)
arg.Page.SetPageSize(2)
volumes, err := suite.volumeService.GetVolumes(arg)
for arg.Page.ContainsMore() {
        volumes, err := suite.volumeService.GetVolumes(arg)
}
```

@rgcostea , I leave it to you to make the call.

rkumpf: Is the server obligated to return all the rows requested?  I wasn't able to discern it from the Nimble REST API documentation.  I've seen in other APIs where the server has discretion over how many rows to return.  For example, if a client requests a million rows, and the server determines it to be excessive, it could elect to return only 100.

Regarding the NextPage function, I wouldn't characterize these updates to the page attributes as "internal", because this function is part of the model within the caller's control.  I see this additional client call as a benefit, while you see it as extra complexity.  Oh well.  :)

vidhutsingh: > Is the server obligated to return all the rows requested? I wasn't able to discern it from the Nimble REST API documentation. I've seen in other APIs where the server has discretion over how many rows to return. For example, if a client requests a million rows, and the server determines it to be excessive, it could elect to return only 100.
> 
> Regarding the NextPage function, I wouldn't characterize these updates to the page attributes as "internal", because this function is part of the model within the caller's control. I see this additional client call as a benefit, while you see it as extra complexity. Oh well. :)

As far as I know the Nimble REST API behavior, it returns all.

rkumpf: If we can confirm the server always returns all requested rows, that simplifies things.  The startRow and endRow values in the response become irrelevant, so we don't have to worry about returning them to the caller.  In that case, I suggest:

1. Remove the PageSize attribute from the Page structure.
2. Implement the SetPageSize function to set StartRow=0 and EndRow=pageSize.
3. In ListFromParams, use the StartRow and EndRow values provided as input, without modification.
4. In ListFromParams, set the TotalRows value in the Page parameter according to the response from the array.  Leave the StartRow and EndRow values unchanged.
5. Change the ContainsMore() function to a NextPage() function that has a dual role of updating the StartRow and EndRow values for the next page and returning a boolean indicating whether there actually is a next page.  For example:

```
func (page *Page) NextPage() boolean {
	if page.StartRow == nil || page.EndRow == nil || page.TotalRows == nil ||
		*page.EndRow >= *page.TotalRows || *page.StartRow >= *page.EndRow {
		return false
	}
	pageSize := *page.EndRow - *page.StartRow
	page.startRow = page.EndRow
	page.SetEndRow(*page.StartRow + pageSize)
	return true
}
```

A caller who chooses to use paging could then use the same logic as the existing implementation, with just one method name change:

```
arg.Page = new(pagination.Page)
arg.Page.SetPageSize(2)
volumes, err := suite.volumeService.GetVolumes(arg)
for arg.Page.NextPage() {
	volumes, err := suite.volumeService.GetVolumes(arg)
}
```

These minor changes would address all three of the concerns I raised earlier:

* Redundant/overlapping attributes in the Page model.
* Input parameters modified with "next page" logic that is hidden inside the Get function and outside the caller's control.
* Input parameters unexpectedly modified with output values.

Again, this assumes the array honors requests for any number of rows.

vidhutsingh: This is exactly what I have implemented, it just that startRow & endRow computation logic has moved from group_mgmt_client.go to pagination.go package. Also, the page field is removed from the struct . 

I accept this change as it makes more sense to have this logic in pagination.go file.

pkg/client/group_mgmt_client.go
```diff
@@ -33,6 +33,7 @@ type DataWrapper struct {
 	StartRow      *int        `json:"startRow,omitempty"`
 	EndRow        *int        `json:"endRow,omitempty"`
 	PageSize      *int        `json:"pageSize,omitempty"`
```
rkumpf, line 35: We'll remove this, right?

pkg/client/group_mgmt_client.go
rkumpf: We can remove this comment.

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,46 @@
 package pagination
 
-type GetPage struct {
+// Page controls enable iterative retrieval of a large data set. This
+// structure is used both on input, to specify the boundaries of a
+// page to retrieve, and on output, to indicate the items actually
```
rkumpf: Change "items" to "number of items", now that we don't need to use the other attributes for output.

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,46 @@
...
-func (page *GetPage) SetEndRow(row int) {
+// SetEndRow - Specify the first row NOT to retrieve in a page of results, using a zero-based index
+// value one larger than the last row desired.
+func (page *Page) SetEndRow(row int) {
 	page.EndRow = &row
 }
-func (page *GetPage) SetTotalRows(row int) {
-	page.TotalRows = &row
+
+// SetPageSize - set the page size for batch processing
```
rkumpf: I suggest:

// Sets the StartRow and EndRow values to request the first page of the specified size.

pkg/param/pagination/pagination.go
```diff
@@ -1,30 +1,46 @@
...
+func (page *Page) SetPageSize(psize int) {
+	zero := 0
+	page.StartRow = &zero
+	page.EndRow = &psize
 }
 
-func (page *GetPage) SetPageSize(row int) {
-	page.PageSize = &row
+// NextPage - Returns true if objects are more than pagesize.
```
rkumpf: I suggest:

  // Updates the StartRow and EndRow values to request the next page of the
  // same size, if more rows are available according to the TotalRows value.
  // Returns true if the values are actually updated.

pkg/service/volume_service_test.go
```diff
@@ -101,6 +101,52 @@ func (suite *VolumeServiceTestSuite) TestGetNonExistentVolumeByID() {
...
+	}
+
+	if len(volumes) != 2 {
+		suite.T().Errorf("TestGetVolumesPagination(): pagination count doesnt match with # of return volumes ")
+		return
+	}
+
+	// Batch processing
+	for arg.Page.NextPage() {
+		volumes, err := suite.volumeService.GetVolumes(arg)
```
rkumpf, line 114: We could avoid duplicate code for getting the first page vs. the remaining pages like this:

	for hasNextPage := true; hasNextPage; hasNextPage = arg.Page.NextPage() {
		_, err := suite.volumeService.GetVolumes(arg)
		if err != nil {
			suite.T().Errorf("TestGetVolumesPagination(): Unable to fetch volumes, err: %v", err.Error())
			return
		}

		if len(volumes) > 2 {
			suite.T().Errorf("TestGetVolumesPagination(): Returned volumes are more than the requested page size")
			return
		}
	}

pkg/param/pagination/pagination.go
rkumpf: Answering my own question...  It seems int is generally preferred.

pkg/param/get_params.go
rkumpf: It will be better to remove the commented code since it is misleading.

rkumpf (commented):
It looks good to me.  Just a few minor items.

suneeth51 (commented):
Hi All, here are some design considerations in Python-SDK to handle pagination in a simple and efficient way so that user don't need to deal with `start_row` and `end_row`.

1) The GET API should be able to specify `limit` (pageSize) and `from_id`. Both are optional parameters.
2) If the GET API is invoked without any parameters, then the SDK should return all the available records from the Array.
3) Whenever user/client/GUI wants to perform pagination at their pace and requirement. This can be performed in the below steps:
       (a) When `limit` (PageSize) is specified, the SDK will return only the limit no. of records.
       (b) To fetch the next  subsequent `limit` (pageSize) no. of records, the caller must pass the ID from last record/item (from the previous step response (a)) as `from_id`.

Please refer this for sample python example:
https://hpe-storage.github.io/nimble-python-sdk/get_started/using/index.html#pagination

pkg/client/group_mgmt_client.go
vidhutsingh: No, let's keep this for now in case if we need this in the future . Anyways this is harmless.

rkumpf: It's harmless as you say, but I'd like to understand your perspective.  This structure is meant to hold the Nimble REST API payload, right?  I don't think the Nimble API defines this attribute, so I'm having trouble imagining when it might become useful.  It seems confusing to keep it.

rkumpf (approved):
Thanks for the updates!

PR #9, opened 2020-09-07
Implemented rpc actions at the service layer 

vidhutsingh:
Feature : Implemented rpc actions at the service layer. 
Details : 
Added consumable methods which include rpc actions for all the supported objects and ops friendly function. Validation of arguments. 
Test:
Added new unit tests 
1. Add volume to volume collection 
2. Restore volume from snapshot 
=== RUN   TestVolumeServiceTestSuite/TestAddVolumeVolcoll
=== RUN   TestVolumeServiceTestSuite/TestOnlineBulkVolumes
=== RUN   TestVolumeServiceTestSuite/TestRestoreVolume
--- PASS: TestVolumeServiceTestSuite (86.72s)
    --- PASS: TestVolumeServiceTestSuite/TestAddVolumeVolcoll (21.75s)
    --- PASS: TestVolumeServiceTestSuite/TestGetNonExistentVolumeByID (20.15s)
    --- PASS: TestVolumeServiceTestSuite/TestGetVolumesPagination (13.62s)
    --- PASS: TestVolumeServiceTestSuite/TestOnlineBulkVolumes (13.97s)
    --- PASS: TestVolumeServiceTestSuite/TestRestoreVolume (17.23s)
PASS
ok  	github.com/hpe-storage/nimble-golang-sdk/pkg/service	86.738s

### DIFF
--- a/pkg/service/active_directory_membership_service.go
+++ b/pkg/service/active_directory_membership_service.go
@@ -25,7 +25,7 @@ func NewActiveDirectoryMembershipService(gs *NsGroupService) *ActiveDirectoryMem
 // GetActiveDirectoryMemberships - method returns a array of pointers of type "ActiveDirectoryMemberships"
 func (svc *ActiveDirectoryMembershipService) GetActiveDirectoryMemberships(params *param.GetParams) ([]*nimbleos.ActiveDirectoryMembership, error) {
 	if params == nil {
-		return nil, fmt.Errorf("error: invalid parameter specified, %v", params)
+		return nil, fmt.Errorf("GetActiveDirectoryMemberships: invalid parameter specified, %v", params)
 	}
 
 	activeDirectoryMembershipResp, err := svc.objectSet.GetObjectListFromParams(params)
@@ -38,7 +38,7 @@ func (svc *ActiveDirectoryMembershipService) GetActiveDirectoryMemberships(param
 // CreateActiveDirectoryMembership - method creates a "ActiveDirectoryMembership"
 func (svc *ActiveDirectoryMembershipService) CreateActiveDirectoryMembership(obj *nimbleos.ActiveDirectoryMembership) (*nimbleos.ActiveDirectoryMembership, error) {
 	if obj == nil {
-		return nil, fmt.Errorf("error: invalid parameter specified, %v", obj)
+		return nil, fmt.Errorf("CreateActiveDirectoryMembership: invalid parameter specified, %v", obj)
 	}
 
 	activeDirectoryMembershipResp, err := svc.objectSet.CreateObject(obj)
@@ -51,7 +51,7 @@ func (svc *ActiveDirectoryMembershipService) CreateActiveDirectoryMembership(obj
 // UpdateActiveDirectoryMembership - method modifies  the "ActiveDirectoryMembership"
 func (svc *ActiveDirectoryMembershipService) UpdateActiveDirectoryMembership(id string, obj *nimbleos.ActiveDirectoryMembership) (*nimbleos.ActiveDirectoryMembership, error) {
 	if obj == nil {
-		return nil, fmt.Errorf("error: invalid parameter specified, %v", obj)
+		return nil, fmt.Errorf("UpdateActiveDirectoryMembership: invalid parameter specified, %v", obj)
 	}
 
 	activeDirectoryMembershipResp, err := svc.objectSet.UpdateObject(id, obj)
@@ -64,7 +64,7 @@ func (svc *ActiveDirectoryMembershipService) UpdateActiveDirectoryMembership(id 
 // GetActiveDirectoryMembershipById - method returns a pointer to "ActiveDirectoryMembership"
 func (svc *ActiveDirectoryMembershipService) GetActiveDirectoryMembershipById(id string) (*nimbleos.ActiveDirectoryMembership, error) {
 	if len(id) == 0 {
-		return nil, fmt.Errorf("error: invalid parameter specified, %v", id)
+		return nil, fmt.Errorf("GetActiveDirectoryMembershipById: invalid parameter specified, %v", id)
 	}
 
 	activeDirectoryMembershipResp, err := svc.objectSet.GetObject(id)
@@ -98,7 +98,7 @@ func (svc *ActiveDirectoryMembershipService) GetActiveDirectoryMembershipByName(
 // DeleteActiveDirectoryMembership - deletes the "ActiveDirectoryMembership"
 func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id string) error {
 	if len(id) == 0 {
-		return fmt.Errorf("error: invalid parameter specified, %s", id)
+		return fmt.Errorf("DeleteActiveDirectoryMembership: invalid parameter specified, %s", id)
 	}
 	err := svc.objectSet.DeleteObject(id)
 	if err != nil {
@@ -110,7 +110,7 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id 
 // RemoveActiveDirectoryMembership - leaves the Active Directory domain.
 //   Required parameters:
 //       id - ID of the active directory.
-//       user - Name of the Activer Directory user with the privilege to leave the domain.
+//       user - "Name of the Active Directory user with the privilege to leave the domain."
 //       password - Password for the Active Directory user.
 
 //   Optional parameters:
@@ -119,7 +119,7 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id 
 func (svc *ActiveDirectoryMembershipService) RemoveActiveDirectoryMembership(id string, user string, password string, force *bool) error {
 
 	if len(id) == 0 || len(user) == 0 || len(password) == 0 {
-		return fmt.Errorf("error: invalid parameter specified id:%v, user:%v, password:%v, force:%v ", id, user, password, force)
+		return fmt.Errorf("RemoveActiveDirectoryMembership: invalid parameter specified id: %v, user: %v, password: %v ", id, user, &password)
 	}
 
 	err := svc.objectSet.Remove(&id, &user, &password, force)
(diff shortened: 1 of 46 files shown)


pkg/service/active_directory_membership_service.go
```diff
@@ -106,3 +106,68 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id
 	}
 	return nil
 }
+
+// RemoveActiveDirectoryMembership - leaves the Active Directory domain.
+//   Required parameters:
+//       id - ID of the active directory.
+//       user - Name of the Activer Directory user with the privilege to leave the domain.
```
rgcostea: typo: Activer

pkg/service/active_directory_membership_service.go
```diff
@@ -106,3 +106,68 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id
...
+//       password - Password for the Active Directory user.
+
+//   Optional parameters:
+//       force - Use this option when there is an error when leaving the domain.
+
+func (svc *ActiveDirectoryMembershipService) RemoveActiveDirectoryMembership(id string, user string, password string, force *bool) error {
+
+	if len(id) == 0 || len(user) == 0 || len(password) == 0 {
+		return fmt.Errorf("error: invalid parameter specified id:%v, user:%v, password:%v, force:%v ", id, user, password, force)
```
rgcostea: add a space after each ":", before %v

rgcostea (approved):
lgtm

Is this all generated?

pkg/service/active_directory_membership_service.go
```diff
@@ -106,3 +106,68 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id
...
+
+func (svc *ActiveDirectoryMembershipService) ReportStatusActiveDirectoryMembership(id string) (*nimbleos.NsADReportStatusReturn, error) {
+
+	if len(id) == 0 {
+		return nil, fmt.Errorf("error: invalid parameter specified id:%v ", id)
+	}
+
+	resp, err := svc.objectSet.ReportStatus(&id)
+	return resp, err
+
```
rkumpf: Remove the blank line.  Also line 157.

pkg/service/array_service.go
```diff
@@ -106,3 +106,48 @@ func (svc *ArrayService) DeleteArray(id string) error {
...
+//   Required parameters:
+//       id - ID of the array to perform failover on.
+
+//   Optional parameters:
+//       force - Initiate failover without performing any precheck.
+
+func (svc *ArrayService) FailoverArray(id string, force *bool) error {
+
+	if len(id) == 0 {
+		return fmt.Errorf("error: invalid parameter specified id:%v, force:%v ", id, force)
```
rkumpf: Is there a reason for this message to include values that weren't part of the validation?  Generally they seem to be the pointer values.  Printing a pointer value with %v will print an address, so we might see "force:0xc00002c008" or "force:<nil>".

pkg/service/active_directory_membership_service.go
rkumpf: This could cause a password to end up in a log file somewhere.  It should be avoided.  Anyway, are we sure a blank password is invalid?

pkg/service/active_directory_membership_service.go
vidhutsingh: thanks

pkg/service/active_directory_membership_service.go
vidhutsingh: sure , will fix it

pkg/service/active_directory_membership_service.go
```diff
@@ -106,3 +106,68 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id
...
+}
+
+// ReportStatusActiveDirectoryMembership - reports the detail status of the Active Directory domain.
+//   Required parameters:
+//       id - ID of the active directory.
+
+func (svc *ActiveDirectoryMembershipService) ReportStatusActiveDirectoryMembership(id string) (*nimbleos.NsADReportStatusReturn, error) {
+
+	if len(id) == 0 {
+		return nil, fmt.Errorf("error: invalid parameter specified id:%v ", id)
```
rkumpf: Would it be helpful for the message to include the function name?  Is there a reason to include the word "error"?

vidhutsingh:
> lgtm
> 
> Is this all generated?

Yes George. Even additional supported functions such as associateVolume and disasscocitateVolume in volume service 
are part of generation only.

pkg/service/active_directory_membership_service.go
```diff
@@ -106,3 +106,68 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id
...
+
+// TestUserActiveDirectoryMembership - tests whether the user exist in the Active Directory. If the user is present, then the user's group and role information is reported.
+//   Required parameters:
+//       id - ID of the Active Directory.
+//       name - Name of the Active Directory user.
+
+func (svc *ActiveDirectoryMembershipService) TestUserActiveDirectoryMembership(id string, name string) (*nimbleos.NsADTestUserReturn, error) {
+
+	if len(id) == 0 || len(name) == 0 {
+		return nil, fmt.Errorf("error: invalid parameter specified id:%v, name:%v ", id, name)
```
rkumpf: It would be nice if we were more specific about which argument is considered invalid, though it might not be worth the extra code to do it.

rkumpf (commented):
Still reviewing....

pkg/service/group_service.go
```diff
@@ -106,3 +106,260 @@ func (svc *GroupService) DeleteGroup(id string) error {
...
+//       srcGroupName - Name of the source group.
+//       srcGroupIp - IP address of the source group.
+//       srcUsername - Username of the source group.
+//       srcPassword - Password of the source group.
+
+//   Optional parameters:
+//       srcPassphrase - Source group encryption passphrase.
+//       skipSecondaryMgmtIp - Skip check for secondary management IP address.
+
+func (svc *GroupService) ValidateMergeGroup(id string, srcGroupName string, srcGroupIp string, srcUsername string, srcPassword string, srcPassphrase *string, skipSecondaryMgmtIp *bool) (*nimbleos.NsGroupMergeReturn, error) {
```
rkumpf, line 265: Some of the function names get a bit difficult to parse.  I wonder, might it be better to remove the noun from the function name, since anyway it is implied by the structure it acts on?

pkg/service/group_service.go
```diff
@@ -106,3 +106,260 @@ func (svc *GroupService) DeleteGroup(id string) error {
...
+
+// GetEulaGroup - get URL to download EULA contents.
+//   Required parameters:
+//       id - ID of the group.
+
+//   Optional parameters:
+//       locale - Locale of EULA contents. Default is en.
+//       format - Format of EULA contents. Default is HTML.
+//       phase - Phase of EULA contents. Default is setup.
+//       force - Flag to force EULA.
```
rkumpf, line 306: The default value is specified for the other parameters.  It would be nice to do so for this one and others elsewhere, if feasible.

pkg/service/active_directory_membership_service.go
rkumpf: Also group_service lines 321 and 364.  Maybe others.

vidhutsingh: sure, thanks

pkg/service/array_service.go
vidhutsingh: The intention was to log all the arguments for the end-user . we could remove it and display the only argument under validation.

rkumpf: Since this is generated code, I realize exceptions to the rule are annoying.  But, I think customers would find logging of passwords to be objectionable (or worse).  Maybe the generator could skip validation of parameters named "password"?

pkg/service/active_directory_membership_service.go
vidhutsingh: Yes we can include function name . wrt 'error' we can go either way but including 'error' makes it more explicit.

pkg/service/active_directory_membership_service.go
vidhutsingh: we are doing validation on 'required parameters' to the function call.

pkg/service/group_service.go
vidhutsingh: It will be a lot more complicated to handle such things in a generator.

pkg/service/group_service.go
vidhutsingh: This is straight coming out from sm object XML file, don't think it is feasible.

pkg/service/array_service.go
vidhutsingh: I agree. will remove 'password' from logging.

rkumpf: Thanks.  Since this message is only generated when an argument is invalid, its purpose seems to be to help identify the invalid argument.  Including non-validated parameters might be more confusing than helpful.  So, I agree with your idea to include only the arguments under validation.

pkg/service/group_service.go
rkumpf: Oh, I was guessing that the generator was putting the verb and noun together.  If the name comes directly from the source, I agree it doesn't seem practical to add logic to break them apart.

pkg/service/group_service.go
rkumpf: I suspected as much.  This comment should be directed upstream.  :)

vidhutsingh:
Thanks @rgcostea  and @rkumpf for reviewing the code. The final changes are merged .

pkg/service/array_service.go
vidhutsingh: It was getting complicated to handle individual 'password' argument from the line so I kept it and now printing only the address of it. That should be fine. So, it will actually log the address of the password, not the actual password.

rgcostea: I find printing the address of the password (or anything for that matter) useless.  Why can't we just skip that?

vidhutsingh: Yes, it should be ignored. But even the current implementation doesn't reveal the password .

Freemarker tool doesn't not have a good programming language syntax even doing a simple thing is complex, the generator process the Action for all the Objects and their Arguments. It can be implemented by selecting the argument 'password' and building logic to ignore it. But it is not worth the effort I feel, do we still need it?

rgcostea: Yes, I feel we should.  We actually do something similar in the java SDK where we look for the keywords password, token, session, etc. and make sure to blank those out.

vidhutsingh: Sure , i will address this issue via a separate bug. thanks

pkg/service/active_directory_membership_service.go
```diff
@@ -110,7 +110,7 @@ func (svc *ActiveDirectoryMembershipService) DeleteActiveDirectoryMembership(id
 // RemoveActiveDirectoryMembership - leaves the Active Directory domain.
 //   Required parameters:
 //       id - ID of the active directory.
-//       user - Name of the Activer Directory user with the privilege to leave the domain.
+//       user - "Name of the Active Directory user with the privilege to leave the domain."
```
rkumpf, line 113: The quotes are odd.  I guess it is an artifact of having to override the upstream message to correct a typo.  Do the upstream messages ever change?

vidhutsingh: There is a bug in the upstream . I will fix the quote in next PR. thanks

PR #11, opened 2020-09-23
Feature : Implemented async job handling

vidhutsingh:
Feature : Implemented asynchronous job handling with REST API call. REST API call returns HTTP code 202 with async job_id . SDK monitors periodically this job_id . It queries the job after every 5 second until default timeout of 300 second. The  async setting works only when sync flag is set at NsGroupService level. 

While working on this there was issue discovered during GetObject() API call in objectset layer which is also fixed. Additional validation of apiVersion is added . 

Test:
Added new test TestAclVolume() (I had to manually stop/start the dsd for verification of async job)
Here is the result :
=== RUN   TestVolumeServiceTestSuite/TestACLVolume
=== RUN   TestVolumeServiceTestSuite/TestAddVolumeVolcoll
=== RUN   TestVolumeServiceTestSuite/TestCloneVolume
=== RUN   TestVolumeServiceTestSuite/TestOnlineBulkVolumes
=== RUN   TestVolumeServiceTestSuite/TestRestoreVolume
--- PASS: TestVolumeServiceTestSuite (116.87s)
    --- PASS: TestVolumeServiceTestSuite/TestACLVolume (18.40s)
    --- PASS: TestVolumeServiceTestSuite/TestAddVolumeVolcoll (16.06s)
    --- PASS: TestVolumeServiceTestSuite/TestCloneVolume (17.00s)
    --- PASS: TestVolumeServiceTestSuite/TestGetNonExistentVolumeByID (13.23s)
    --- PASS: TestVolumeServiceTestSuite/TestGetVolumesPagination (13.19s)
    --- PASS: TestVolumeServiceTestSuite/TestOnlineBulkVolumes (14.24s)
    --- PASS: TestVolumeServiceTestSuite/TestRestoreVolume (24.77s)
PASS
ok  	github.com/hpe-storage/nimble-golang-sdk/pkg/service	116.890s 

### DIFF
--- a/pkg/client/access_control_records_object_set.go
+++ b/pkg/client/access_control_records_object_set.go
@@ -1,7 +1,5 @@
 // Copyright 2020 Hewlett Packard Enterprise Development LP
 
-// create true
-
 package client
 
 import (
--- a/pkg/client/access_control_records_object_set.go
+++ b/pkg/client/access_control_records_object_set.go
@@ -8,7 +8,6 @@ import (
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"reflect"
-	"strings"
 )
 
 // Manage access control records for volumes.
@@ -25,14 +24,6 @@ type AccessControlRecordObjectSet struct {
 func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
 	resp, err := objectSet.Client.Post(accessControlRecordPath, payload, &nimbleos.AccessControlRecord{})
 	if err != nil {
-		//process http code 202
-		if strings.Contains(err.Error(), "status (202)") {
-			if resp != nil {
-				ID := resp.(string)
-				// Get object
-				return objectSet.GetObject(ID)
-			}
-		}
 		return nil, err
 	}
 
--- a/pkg/client/access_control_records_object_set.go
+++ b/pkg/client/access_control_records_object_set.go
@@ -1,5 +1,7 @@
 // Copyright 2020 Hewlett Packard Enterprise Development LP
 
+// create true
+
 package client
 
 import (
@@ -8,6 +10,7 @@ import (
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"reflect"
+	"strings"
 )
 
 // Manage access control records for volumes.
@@ -24,6 +27,14 @@ type AccessControlRecordObjectSet struct {
 func (objectSet *AccessControlRecordObjectSet) CreateObject(payload *nimbleos.AccessControlRecord) (*nimbleos.AccessControlRecord, error) {
 	resp, err := objectSet.Client.Post(accessControlRecordPath, payload, &nimbleos.AccessControlRecord{})
 	if err != nil {
+		//process http code 202
+		if strings.Contains(err.Error(), "status (202)") {
+			if resp != nil {
+				ID := resp.(string)
+				// Get object
+				return objectSet.GetObject(ID)
+			}
+		}
 		return nil, err
 	}
 
@@ -46,7 +57,7 @@ func (objectSet *AccessControlRecordObjectSet) DeleteObject(id string) error {
 
 // GetObject returns a AccessControlRecord object with the given ID
 func (objectSet *AccessControlRecordObjectSet) GetObject(id string) (*nimbleos.AccessControlRecord, error) {
-	resp, err := objectSet.Client.Get(accessControlRecordPath, id, nimbleos.AccessControlRecord{})
+	resp, err := objectSet.Client.Get(accessControlRecordPath, id, &nimbleos.AccessControlRecord{})
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -9,6 +9,7 @@ import (
 	"github.com/go-resty/resty/v2"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"strings"
 	"time"
 )
 
@@ -125,7 +126,10 @@ func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder
 			// extract error message
 			id, err := processAsyncResponse(client, response.Body())
 			if id != nil {
-				return client.Get(path, id.(string), respHolder)
+				// action rpc may contains different path.
+				// extract Get uri path from original path.
+				newPath := strings.Split(path, "/")
+				return client.Get(newPath[0], id.(string), respHolder)
 			} else {
 				return nil, err
 			}
@@ -162,7 +166,10 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
 			// extract error message
 			id, err := processAsyncResponse(client, response.Body())
 			if id != nil {
-				return client.Get(path, id.(string), respHolder)
+				// action rpc may contains different path.
+				// extract Get uri path from original path.
+				newPath := strings.Split(path, "/")
+				return client.Get(newPath[0], id.(string), respHolder)
 			} else {
 				return nil, err
 			}
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -59,7 +59,7 @@ type Argument struct {
 }
 
 // NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string, sync bool) (*GroupMgmtClient, error) {
+func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
 	// Create GroupMgmt Client
 	client := resty.New()
 	client.SetTLSClientConfig(&tls.Config{
@@ -68,7 +68,7 @@ func NewClient(ipAddress, username, password, apiVersion string, sync bool) (*Gr
 	groupMgmtClient := &GroupMgmtClient{
 		URL:       fmt.Sprintf("https://%s:5392/%s", ipAddress, apiVersion),
 		Client:    client,
-		WaitOnJob: sync,
+		WaitOnJob: waitOnJobs,
 	}
 
 	// Get session token
@@ -125,21 +125,7 @@ func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder
 		// http code 202, handle async job
 		if response.StatusCode() == 202 {
 			// extract error message
-			errResp, _ := unwrapError(response.Body())
-			if client.WaitOnJob { // check sync flag
-				unwrapMessage := &ErrorResponse{}
-				err = json.Unmarshal(response.Body(), unwrapMessage)
-				if err != nil {
-					return nil, err
-				}
-				ids, err := handleAsyncJob(unwrapMessage, client)
-
-				if err != nil {
-					return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), err.Error())
-				}
-				return ids, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
-			}
-			return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
+			return processAsyncResponse(client, response.Body())
 		}
 		return unwrap(response.Body(), respHolder)
 	}
@@ -171,21 +157,7 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
 		// http code 202, handle async job
 		if response.StatusCode() == 202 {
 			// extract error message
-			errResp, _ := unwrapError(response.Body())
-			if client.WaitOnJob { // check sync flag
-				unwrapMessage := &ErrorResponse{}
-				err = json.Unmarshal(response.Body(), unwrapMessage)
-				if err != nil {
-					return nil, err
-				}
-				ids, err := handleAsyncJob(unwrapMessage, client)
-
-				if err != nil {
-					return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), err.Error())
-				}
-				return ids, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
-			}
-			return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
+			return processAsyncResponse(client, response.Body())
 		}
 		return unwrap(response.Body(), respHolder)
 	}
@@ -424,6 +396,26 @@ func unwrapError(body []byte) (string, error) {
 	return errResp, nil
 }
 
+// processAsyncResponse: process http code 202 response
+func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, error) {
+	errResp, _ := unwrapError(body)
+	if client.WaitOnJob { // check sync flag
+		unwrapMessage := &ErrorResponse{}
+		err := json.Unmarshal(body, unwrapMessage)
+		if err != nil {
+			return nil, err
+		}
+		ids, err := handleAsyncJob(unwrapMessage, client)
+
+		if err != nil {
+			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
+		}
+		return ids, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+	}
+	return nil, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+}
+
+//handleAsyncJob : extract the job_id and monitor it periodically until job completion or timed out
 func handleAsyncJob(resp *ErrorResponse, client *GroupMgmtClient) (string, error) {
 	// filter list of async job id from the message
 
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -9,7 +9,6 @@ import (
 	"github.com/go-resty/resty/v2"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
-	"strings"
 	"time"
 )
 
@@ -416,7 +415,7 @@ func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, er
 		}
 		var jobId string
 		for _, msg := range unwrapMessage.Messages {
-			if strings.Compare(msg.Code, smAsyncJobId) == 0 {
+			if msg.Code == smAsyncJobId {
 				jobId = msg.Arguments.JobId
 			}
 		}
@@ -428,7 +427,7 @@ func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, er
 		if err != nil {
 			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
 		}
-		return id, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+		return id, nil
 	}
 	return nil, fmt.Errorf("http response error: status (202), messages: %v", errResp)
 }
@@ -454,8 +453,7 @@ func waitForJobResult(jobId string, client *GroupMgmtClient) (interface{}, error
 			}
 
 		case <-timeoutChan:
-			return nil, fmt.Errorf("waitForJobResult: wait on jobs in progress has timed out. Following %v jobs didnt complete on time.", jobId)
+			return nil, fmt.Errorf("waitForJobResult: job with ID %v timed out after %v seconds.", jobId, jobTimeout)
 		}
 	}
-	return nil, nil
 }
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -14,13 +14,12 @@ import (
 )
 
 const (
-	groupURIFmt        = "https://%s:5392/%s"
-	clientTimeout      = time.Second * 60 // 1 Minute
-	maxLoginRetries    = 5
-	retrySleepDuration = 2                 // Second
-	jobTimeout         = time.Second * 300 // 5 Minute
-	retryFrequency     = 5 * time.Second   // Second
-	smAsyncJobId       = "SM_async_job_id"
+	groupURIFmt     = "https://%s:5392/%s"
+	clientTimeout   = time.Second * 60 // 1 Minute
+	maxLoginRetries = 5
+	jobTimeout      = time.Second * 300 // 5 Minute
+	jobPollInterval = 5 * time.Second   // Second
+	smAsyncJobId    = "SM_async_job_id"
 )
 
 // GroupMgmtClient :
@@ -415,35 +414,31 @@ func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, er
 		if err != nil {
 			return nil, err
 		}
-		ids, err := handleAsyncJob(unwrapMessage, client)
+		var jobId string
+		for _, msg := range unwrapMessage.Messages {
+			if strings.Compare(msg.Code, smAsyncJobId) == 0 {
+				jobId = msg.Arguments.JobId
+			}
+		}
+		if len(jobId) == 0 {
+			return nil, fmt.Errorf("http response error: status (202), failed to get the job id")
+		}
 
+		id, err := waitForJobResult(jobId, client)
 		if err != nil {
 			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
 		}
-		return ids, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+		return id, fmt.Errorf("http response error: status (202), messages: %v", errResp)
 	}
 	return nil, fmt.Errorf("http response error: status (202), messages: %v", errResp)
 }
 
-//handleAsyncJob : extract the job_id and monitor it periodically until job completion or timed out
-func handleAsyncJob(resp *ErrorResponse, client *GroupMgmtClient) (string, error) {
-	// filter list of async job id from the message
+//waitForJobResult : it monitors jobId periodically until job completion or timed out
+func waitForJobResult(jobId string, client *GroupMgmtClient) (interface{}, error) {
 
-	var objectId string
-	var jobId string
-
-	for _, msg := range resp.Messages {
-		if strings.Compare(msg.Code, smAsyncJobId) == 0 {
-			jobId = msg.Arguments.JobId
-		}
-	}
-
-	if len(jobId) == 0 {
-		return jobId, fmt.Errorf("handleAsyncJob: failed to get the job id")
-	}
 	// Loop over job ids periodically unitl 300 sec timeout or unitl completion of jobs.
-	intervalChan := time.Tick(retryFrequency) // control the fequency of GetObject() API call.
-	timeoutChan := time.After(jobTimeout)     // timeout setting, 300 Seconds
+	intervalChan := time.Tick(jobPollInterval) // control the fequency of GetObject() API call.
+	timeoutChan := time.After(jobTimeout)      // timeout setting, 300 Seconds
 	for {
 		select {
 		case <-intervalChan:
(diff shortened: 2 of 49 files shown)


pkg/client/access_control_records_object_set.go
```diff
@@ -1,5 +1,7 @@
 // Copyright 2020 Hewlett Packard Enterprise Development LP
 
+// create true
+
```
rgcostea: Why are these comments in ever file?

pkg/client/access_control_records_object_set.go
```diff
@@ -24,6 +27,14 @@ type AccessControlRecordObjectSet struct {
...
 	resp, err := objectSet.Client.Post(accessControlRecordPath, payload, &nimbleos.AccessControlRecord{})
 	if err != nil {
+		//process http code 202
+		if strings.Contains(err.Error(), "status (202)") {
+			if resp != nil {
+				ID := resp.(string)
+				// Get object
+				return objectSet.GetObject(ID)
+			}
+		}
```
rgcostea: We should not be repeating this logic in every ObjectSet.  Please push the job handling down into the GroupMgmtClient.

pkg/client/group_mgmt_client.go
```diff
@@ -44,20 +48,27 @@ type ErrorResponse struct {
...
+	Arguments Argument `json:"arguments,omitempty"`
+}
+
+type Argument struct {
+	JobId string `json:"job_id,omitempty"`
 }
 
 // NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string) (*GroupMgmtClient, error) {
+func NewClient(ipAddress, username, password, apiVersion string, sync bool) (*GroupMgmtClient, error) {
```
rgcostea: Rename sync to waitOnJobs.  It's not obvious that that is what sync means.

pkg/client/group_mgmt_client.go
```diff
@@ -136,12 +167,32 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
...
+				ids, err := handleAsyncJob(unwrapMessage, client)
+
+				if err != nil {
+					return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), err.Error())
+				}
+				return ids, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
+			}
+			return nil, fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)
+		}
+		return unwrap(response.Body(), respHolder)
```
rgcostea: This looks like common code that we can move into a function.

rgcostea (changes requested):
Please address review comments.  Thanks.

pkg/client/access_control_records_object_set.go
vidhutsingh: It was ftl debug print :) , will remove it . thanks

pkg/client/access_control_records_object_set.go
vidhutsingh: This cannot be done as the objectSet type is known only to object set layer . Example : ID could be of volume or it could be of snapshot .

pkg/client/group_mgmt_client.go
vidhutsingh: sure . thanks

pkg/client/group_mgmt_client.go
vidhutsingh: sure .

pkg/client/access_control_records_object_set.go
rgcostea: Yes, this can be done.  The Post function takes the expected response type as an interface:

`func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder interface{})`

So why not check the http response status within that function?  If you have a 202, then you call Get from the client after the job completes:

`func (client *GroupMgmtClient) Get(path string, id string, intf interface{})`

vidhutsingh: I see . Yes , it seems possible, the only concern is "path" even that is the same for both POST/PUT and GET . thanks @rgcostea

pkg/client/group_mgmt_client.go
```diff
@@ -157,7 +162,12 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
 		// http code 202, handle async job
 		if response.StatusCode() == 202 {
 			// extract error message
-			return processAsyncResponse(client, response.Body())
+			id, err := processAsyncResponse(client, response.Body())
+			if id != nil {
+				return client.Get(path, id.(string), respHolder)
+			} else {
+				return nil, err
+			}
```
rgcostea: Common code again.  Can this be moved to processAsyncResponse?

pkg/client/group_mgmt_client.go
```diff
@@ -6,25 +6,29 @@ import (
...
+	"strings"
+	"time"
 )
 
 const (
 	groupURIFmt        = "https://%s:5392/%s"
 	clientTimeout      = time.Second * 60 // 1 Minute
 	maxLoginRetries    = 5
-	retrySleepDuration = 2 // Seconds
+	retrySleepDuration = 2                 // Second
```
rkumpf: This constant does not appear to be used.  Remove it?

pkg/client/group_mgmt_client.go
```diff
@@ -6,25 +6,29 @@ import (
...
+	jobTimeout         = time.Second * 300 // 5 Minute
+	retryFrequency     = 5 * time.Second   // Second
```
rkumpf: I suggest the name jobPollInterval.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+			return nil, err
+		}
+		ids, err := handleAsyncJob(unwrapMessage, client)
+
+		if err != nil {
+			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
+		}
+		return ids, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+	}
+	return nil, fmt.Errorf("http response error: status (202), messages: %v", errResp)
```
rkumpf, line 439: I would be interested to see example code for a caller who specifies WaitOnJob=false and polls on the job ID themselves.  It doesn't look very simple to extract the job ID from this error, or even to discern that it is not actually an error.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+// processAsyncResponse: process http code 202 response
+func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, error) {
+	errResp, _ := unwrapError(body)
+	if client.WaitOnJob { // check sync flag
+		unwrapMessage := &ErrorResponse{}
+		err := json.Unmarshal(body, unwrapMessage)
+		if err != nil {
+			return nil, err
+		}
+		ids, err := handleAsyncJob(unwrapMessage, client)
```
rkumpf: Should "ids" be "id"?  It might be helpful to use "jobId" instead.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+
+		if err != nil {
+			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
+		}
+		return ids, fmt.Errorf("http response error: status (202), messages: %v", errResp)
```
rkumpf: What is the purpose of returning an error on success?  Isn't the normal pattern to look at the result only if the error is nil?  If we need it, we could clarify the behavior in the function description.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+	// Loop over job ids periodically unitl 300 sec timeout or unitl completion of jobs.
+	intervalChan := time.Tick(retryFrequency) // control the fequency of GetObject() API call.
+	timeoutChan := time.After(jobTimeout)     // timeout setting, 300 Seconds
+	for {
+		select {
+		case <-intervalChan:
+
+			job, err := client.GetJobObjectSet().GetObject(jobId)
+			if err != nil {
+				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
```
rkumpf, line 454: How about this?

  fmt.Println("Failed to get job info for ID %s, err : %s", jobId, err.Error())

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
 	}
 	return errResp, nil
 }
+
+// processAsyncResponse: process http code 202 response
+func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, error) {
```
rkumpf, line 415: It looks like we are returning a job ID.  Should the return type be string?

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+				if strings.Compare(string(*job.State), string(*nimbleos.NsJobStatusDone)) == 0 {
+					return objectId, nil
+				}
+			}
+
+		case <-timeoutChan:
+			return objectId, fmt.Errorf("handleAsyncJob: wait on jobs in progress has timed out. Following %v jobs didnt complete on time.", jobId)
+		}
+	}
+	return objectId, nil
```
rkumpf: Is it possible to reach here?

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+			} else {
+				objectId = *job.ObjectId
+				if strings.Compare(string(*job.State), string(*nimbleos.NsJobStatusDone)) == 0 {
+					return objectId, nil
+				}
+			}
+
+		case <-timeoutChan:
+			return objectId, fmt.Errorf("handleAsyncJob: wait on jobs in progress has timed out. Following %v jobs didnt complete on time.", jobId)
```
rkumpf: I suggest:

  fmt.Errorf("handleAsyncJob: Job with ID %v timed out after %v seconds.", jobId, jobTimeout)

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+			if err != nil {
+				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+			} else {
+				objectId = *job.ObjectId
+				if strings.Compare(string(*job.State), string(*nimbleos.NsJobStatusDone)) == 0 {
+					return objectId, nil
+				}
+			}
+
+		case <-timeoutChan:
```
rkumpf, line 462: What is the purpose of returning the objectId in this case?  It is not useful unless the job completes, and someone who checks if the job has completed can get the objectId that way.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+		return ids, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+	}
+	return nil, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+}
+
+//handleAsyncJob : extract the job_id and monitor it periodically until job completion or timed out
+func handleAsyncJob(resp *ErrorResponse, client *GroupMgmtClient) (string, error) {
+	// filter list of async job id from the message
+
+	var objectId string
```
rkumpf: Based on my other comments below, maybe we only need this variable in scope at line 455.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
+		}
+		return ids, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+	}
+	return nil, fmt.Errorf("http response error: status (202), messages: %v", errResp)
+}
+
+//handleAsyncJob : extract the job_id and monitor it periodically until job completion or timed out
+func handleAsyncJob(resp *ErrorResponse, client *GroupMgmtClient) (string, error) {
```
rkumpf: The purpose of this function seems a bit jumbled.  Maybe it would help if we move the job ID extraction back into processAsyncResponse at line 418 (since it is more like processing the response than handling the job) and rename this function to:

func waitForJobResult(jobId string, client *GroupMgmtClient) (string, error)

pkg/service/group_service_factory.go
```diff
@@ -60,8 +62,11 @@ type NsGroupService struct {
 }
 
 // NewNsGroupService - initializes NsGroupService
-func NewNsGroupService(ip string, username string, password string) (gs *NsGroupService, err error) {
-	client, err := client.NewClient(ip, username, password, "v1")
+func NewNsGroupService(ip, username, password, apiVersion string, synchronous bool) (gs *NsGroupService, err error) {
+	if strings.Compare(apiVersion, "v1") != 0 {
```
rkumpf: Simpler check?

  if apiVersion != "v1" {

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +405,62 @@ func unwrapError(body []byte) (string, error) {
...
+	for {
+		select {
+		case <-intervalChan:
+
+			job, err := client.GetJobObjectSet().GetObject(jobId)
+			if err != nil {
+				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+			} else {
+				objectId = *job.ObjectId
+				if strings.Compare(string(*job.State), string(*nimbleos.NsJobStatusDone)) == 0 {
```
rkumpf: Simpler check?

  if string(*job.State) == string(*nimbleos.NsJobStatusDone) {

pkg/client/group_mgmt_client.go
vidhutsingh: sure.

pkg/client/group_mgmt_client.go
vidhutsingh: Ok

pkg/client/group_mgmt_client.go
vidhutsingh: we arent extracting job id from error. job id is returned as second argument.

pkg/client/group_mgmt_client.go
vidhutsingh: yes it should be id .

pkg/client/group_mgmt_client.go
vidhutsingh: Lot many things can go wrong while polling like a timeout or GetObject() fails , unmarhsalling error etc.

pkg/client/group_mgmt_client.go
vidhutsingh: we can return nil here .

pkg/client/group_mgmt_client.go
vidhutsingh: Yes , it is due to the refactoring of common code . Anyways i will move jobiD extraction to processAsyncResponse.

pkg/service/group_service_factory.go
vidhutsingh: Make sense.

pkg/client/group_mgmt_client.go
vidhutsingh: Not much of a common code otherwise we will have to pass "path", respHolder additional argument to processAsyncResponse().

pkg/client/group_mgmt_client.go
vidhutsingh: done

pkg/client/group_mgmt_client.go
vidhutsingh: Yes.

pkg/client/group_mgmt_client.go
```diff
@@ -6,25 +6,28 @@ import (
...
 
 const (
-	groupURIFmt        = "https://%s:5392/%s"
-	clientTimeout      = time.Second * 60 // 1 Minute
-	maxLoginRetries    = 5
-	retrySleepDuration = 2 // Seconds
+	groupURIFmt     = "https://%s:5392/%s"
+	clientTimeout   = time.Second * 60 // 1 Minute
+	maxLoginRetries = 5
+	jobTimeout      = time.Second * 300 // 5 Minute
```
rkumpf, line 20: Should we consider allowing the client program to specify/override this value?

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +404,58 @@ func unwrapError(body []byte) (string, error) {
...
+	errResp, _ := unwrapError(body)
+	if client.WaitOnJob { // check sync flag
+		unwrapMessage := &ErrorResponse{}
+		err := json.Unmarshal(body, unwrapMessage)
+		if err != nil {
+			return nil, err
+		}
+		var jobId string
+		for _, msg := range unwrapMessage.Messages {
+			if strings.Compare(msg.Code, smAsyncJobId) == 0 {
```
rkumpf: Use == here?

pkg/client/group_mgmt_client.go
rkumpf: Can you please elaborate?

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +404,58 @@ func unwrapError(body []byte) (string, error) {
...
+				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+			} else {
+				var objectId = *job.ObjectId
+				if string(*job.State) == string(*nimbleos.NsJobStatusDone) {
+					return objectId, nil
+				}
+			}
+
+		case <-timeoutChan:
+			return nil, fmt.Errorf("waitForJobResult: wait on jobs in progress has timed out. Following %v jobs didnt complete on time.", jobId)
```
rkumpf: I suggested a message change earlier, but I realize now I should have been more clear in writing the entire statement rather than just the error part:

  return nil, fmt.Errorf("waitForJobResult: Job with ID %v timed out after %v seconds.", jobId, jobTimeout)

pkg/client/group_mgmt_client.go
```diff
@@ -44,20 +47,27 @@ type ErrorResponse struct {
...
+	Arguments Argument `json:"arguments,omitempty"`
+}
+
+type Argument struct {
+	JobId string `json:"job_id,omitempty"`
 }
 
 // NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string) (*GroupMgmtClient, error) {
+func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
```
rkumpf, line 61: Rather than requiring the waitForJobs option to be specified here, would it make sense to set a default value and provide another function to change it?  That pattern could open the door to additional configurability.  For example the values at lines 17-21 could be overridden.

pkg/client/group_mgmt_client.go
```diff
@@ -384,3 +404,58 @@ func unwrapError(body []byte) (string, error) {
...
+		}
+		if len(jobId) == 0 {
+			return nil, fmt.Errorf("http response error: status (202), failed to get the job id")
+		}
+
+		id, err := waitForJobResult(jobId, client)
+		if err != nil {
+			return nil, fmt.Errorf("http response error: status (202), messages: %v", err.Error())
+		}
+		return id, fmt.Errorf("http response error: status (202), messages: %v", errResp)
```
rkumpf: Sorry, I didn't understand the response to my prior comment at this line.  If we reach this statement, the job has completed successfully.  Why return an error in addition to the object ID?

pkg/client/group_mgmt_client.go
vidhutsingh: I am not sure if this should be supported with this version of the release , this may be required for more advanced users or if we see the need for this or any issue. The sdk is consumed directly as source code so user-specific settings can be done manually to the source file unlike other sdk deliverables (binaries or share lib). 
@rgcostea , what is your view on this ?

pkg/client/group_mgmt_client.go
vidhutsingh: ok

pkg/client/group_mgmt_client.go
vidhutsingh: Made the change . Also yes id,nil should be returned.

pkg/client/group_mgmt_client.go
vidhutsingh: It should be nil

pkg/client/group_mgmt_client.go
vidhutsingh: I think for this release it is ok. please check the above comment in the same context.

pkg/client/group_mgmt_client.go
vidhutsingh: done

pkg/client/group_mgmt_client.go
vidhutsingh: we dont need to return error . returnig nil .

vidhutsingh: It is fixed . We dont need to return error. this comment is appearing at multiplaces and have already replied .

vidhutsingh:
> Please address review comments. Thanks.

thanks for review. The code review changes have been done.

pkg/client/group_mgmt_client.go
rgcostea: I like the idea of letting the user set the timeout actually.

pkg/client/group_mgmt_client.go
rgcostea, line 61: Sorry for the late comment here.  Can we change the constructor to use the functional options pattern: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Thanks

vidhutsingh: @rgcostea , can we take this as an enhancement and close this pull request?

pkg/client/group_mgmt_client.go
vidhutsingh: Okay, we can track this via different PR as commented below.

pkg/client/group_mgmt_client.go
rkumpf: I'm still wondering about this.  I see at line 132 we must cast the return value to a string.  What is the purpose for returning a generic interface{}?

pkg/client/group_mgmt_client.go
rkumpf: We provide the option for the user to set WaitOnJob to false.  The use model seems to be that the caller doesn't want to block while the job completes, and they would want to check for job completion asynchronously.  To do this, the caller will require the job ID.  I'd like to understand how the caller can access the job ID in this case.

volume_service_test.go always sets WaitOnJobs to true when creating the group service.  It would be great to see an example where WaitOnJobs is set to false.  Call an interface that returns a 202 status, get the job ID in return, and then check the job status using a separate call into the SDK.

pkg/client/group_mgmt_client.go
rkumpf: It seems like we are using the job ID as a verb, which makes it difficult for me to understand the message.

pkg/client/group_mgmt_client.go
vidhutsingh: Sorry I missed this comment earlier. We don’t really need to support this use case since customer can always invoke GetJobs services to get all the services  he wants to monitor separately.

pkg/client/group_mgmt_client.go
vidhutsingh: We can fix this later. In next PR

pkg/client/group_mgmt_client.go
rkumpf: How does the user know which job is the one that he wants to monitor?  I.e., the one that was created from this request.

pkg/client/group_mgmt_client.go
vidhutsingh: It’s better to return nil and handle nil check rather empty string check .

pkg/client/group_mgmt_client.go
vidhutsingh: It contains object I’d as well

pkg/client/group_mgmt_client.go
rkumpf: Return a *string, then?

pkg/client/group_mgmt_client.go
rkumpf: How does the object ID in the job help me identify the job that was created from my POST request?

rgcostea: Roger is right here.  We should provide an example of how a client would get the job_id off of the error response and then use it in GetJobs for subsequent monitoring.

vidhutsingh: I should have been more explicit. If the user selects sync=false and the API response is HTTP Code 202 , in that case we return an error to the caller and the error message has job_id in it . The user has to extract the job_id and monitors it. 
Here is an example of error message :
**http response error: status (202), messages: {Code:SM_async_job_id Text:A job has been created, use ID 280695c517be8bd20b000000000000000000000011 to track its status. Severity:error Arguments:{JobId:280695c517be8bd20b000000000000000000000011}}**

Moreover, he can also invoke GetJob() service separately to monitor the respective job. the response contains job_id, name , operation type , object_id etc . it has enough information to filter out the required job.

pkg/client/group_mgmt_client.go
vidhutsingh: There are multiple ways to do the same thing. is there any issue with interface here ?

rkumpf: I think interface{} is for when you don't know what type will be returned.  If you know and you aren't telling me, well, that just seems kind of mean.  :)

Really though, I feel the best approach is to make the return type a string and to change the logic at line 128/158 to check for err == nil instead of id != nil.

pkg/client/group_mgmt_client.go
rkumpf: The error message comes back in string form, right?  It doesn't look so easy to extract the job ID.

How does one determine that the "error" really just means the response will be asynchronous?  It seems unintuitive, since it isn't really an error condition.

In general is there a way for the calling code to reliably determine the HTTP status code in an error case?  For example, how should a long-lived client application handle an auth token expiration?

vidhutsingh: We don’t have concept of exception here with golang and in fact with other sdk it generates exception and user has to handle it. It is not intuitive but if user is interested in monitoring the job by himself he can extract the job ID from error string and it is not difficult

pkg/client/group_mgmt_client.go
vidhutsingh: There is no hard rule that you need use interface only when you don't know the concrete type. you can use interface to write clean code . Consider this that you are returning string . You will have to write if st != "" {} such code . Here I am using interface to pass either string or nil without using a *string or string . In the caller i just need to check if it is nil or not . Clean code.

vidhutsingh: Here is the call stack :
processAsyncResponse(), calls waitForJobResults() which further calls GetJobObjectSet() which returns interface  type . Here in this specific scenario, it is easy to pass interface type in the call stack . Finally, where it is needed to use just need to convert it to a concrete type . 

Also interface let you write clean code and you don't have to juggle with *string or string . 

Here is a sample code 
```
func mystr(s interface{}) {
	if s == nil {
		fmt.Println(nil)
	
	}else {
		fmt.Println(s)
	}
}

func main() {

	var sptr *string
	mystr(sptr)

	sval :="a"
	mystr(sval)
}
```

pkg/client/group_mgmt_client.go
rgcostea: I believe @rkumpf is drawing attention to the fact that the current error we report makes it difficult to determine you have an asynchronous job that you now have to monitor.  The Nimble REST API will return an error in this form:

```
// ErrorResponse is a serializer struct for representing a valid JSON API errors payload.
type ErrorResponse struct {
	Messages []*Message json:"messages"
}
// Message is an Error implementation as well as an implementation of the JSON API error object.
type Message struct {
	Code string json:"code,omitempty"
	Text string json:"text,omitempty"
}
```

And our error messages are reported in this form:

`fmt.Errorf("http response error: status (%d), messages: %v", response.StatusCode(), errResp)`

in order to fit within the Error interface that is provided by golang.  This is a valid concern and worthy of improvement.  There is a lot of discussion on this topic.  Here's one link worth looking into: https://middlemost.com/failure-is-your-domain/.  However, I feel our current implementation will suffice for now.  Can we file an issue to track this as an area of improvement in a future PR?  Thanks.

vidhutsingh: sure thanks

vidhutsingh:
I am blocked due this PR as any new code changes will append to this PR. If there are no further comments, pls approve this PR.

rgcostea (approved):

pkg/client/group_mgmt_client.go
rkumpf: Interesting, I got an email when you added a comment saying you accept the change, but not when you edited the comment as above.

In the call stack you reference, GetJobObjectSet().GetObject() actually returns a *Job, not an interface{}.  The ObjectId is then extracted as a *string.

Both processAsyncResponse() and waitForJobResults() return an interface{} and an error, and I think both should be changed to return a string instead.  If the error is not nil, the string should be ignored, so there is no need to worry about the empty string as a special value.

The example code above is not compelling to me, because you don't explain what it means if the caller passes in a.non-string value.

vidhutsingh: empty interfaces{} accepts any type of value. 
Moreover, this is a minor comment and shouldn't hold back feature merge.

rkumpf: I agree it is a minor comment.  I was under the impression that another commit was coming, based on the comment I received in email.  I didn't know to look for an edited comment.

PR #12, opened 2020-09-29
Feature : Automatic token refresh

vidhutsingh:
Feature : Implement automatic session token refresh if the token expires while running a long job. This will enhance the user experience. 

Details : 
SDK internally now check for HTTP code > 400 and error message "SM_http_unauthorized" while executing CRUD operations . The CRUD operations are executed after refreshing the session token. 

Added a new function to delete the session token as it was missing with SDK. 

Unit Test:
I've written a custom test code to validate this since session tokens expire in 1801 seconds. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # HPE Nimble Storage SDK in Go Programming Language
-This is a Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client library for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
+This is a Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
 
 The SDK provides a Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # HPE Nimble Storage SDK in Go Programming Language
-This is a Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
+This is Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
 
 The SDK provides a Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # HPE Nimble Storage SDK in Go Programming Language
-This is Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
+This is the Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
 
-The SDK provides a Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
+The SDK provides Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
 
 # Synopsis
 
--- a/README.md
+++ b/README.md
@@ -1,1 +1,22 @@
-# nimble-golang-sdk
+# HPE Nimble Storage SDK in Go Programming Language
+This is a Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client library for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
+
+The SDK provides a Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
+
+# Synopsis
+
+
+# Requirements
+
+
+# Contributing
+
+Contributing guidelines are available in the [official documentation](https://hpe-storage.github.io/nimble-golang-sdk/legal/contributing/index.html).
+
+# Community
+
+Join the HPE DEV slack by signing up at [slack.hpedev.io](https://slack.hpedev.io). HPE employees may sign in direct at [hpedev.slack.com](https://hpedev.slack.com) with a valid HPE address. We hang out in `#NimbleStorage`.
+
+# License
+
+See [LICENSE](https://hpe-storage.github.io/nimble-golang-sdk/legal/license/index.html)
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -147,7 +147,7 @@ func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder
 			// retry the ops
 			return client.Post(path, payload, respHolder)
 		}
-		return processError(response.StatusCode(), response.Body())
+		return nil,processError(response.StatusCode(), response.Body())
 	}
 	return processResponse(client, response, path, respHolder)
 }
@@ -178,7 +178,7 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
 			// retry the ops
 			return client.Put(path, id, payload, respHolder)
 		}
-		return processError(response.StatusCode(), response.Body())
+		return nil,processError(response.StatusCode(), response.Body())
 	}
 	return processResponse(client, response, path, respHolder)
 }
@@ -212,7 +212,7 @@ func (client *GroupMgmtClient) Get(path string, id string, respHolder interface{
 			// retry the ops
 			return client.Get(path, id, respHolder)
 		}
-		return processError(response.StatusCode(), response.Body())
+		return nil,processError(response.StatusCode(), response.Body())
 	}
 	return processResponse(client, response, path, respHolder)
 }
@@ -240,8 +240,7 @@ func (client *GroupMgmtClient) Delete(path string, id string) error {
 			// retry the ops
 			return client.Delete(path, id)
 		}
-		_, err = processError(response.StatusCode(), response.Body())
-		return err
+		return processError(response.StatusCode(), response.Body())
 	}
 	_, err = processResponse(client, response, path, nil)
 	return err
@@ -341,7 +340,7 @@ func (client *GroupMgmtClient) listPost(
 			// retry the ops
 			return client.listPost(path, payload, queryParams, params)
 		}
-		return processError(response.StatusCode(), response.Body())
+		return nil,processError(response.StatusCode(), response.Body())
 	}
 
 	if params != nil && params.Page != nil {
@@ -382,7 +381,7 @@ func (client *GroupMgmtClient) listGet(
 			// retry the ops
 			return client.listGet(path, queryParams, params)
 		}
-		return processError(response.StatusCode(), response.Body())
+		return nil,processError(response.StatusCode(), response.Body())
 	}
 
 	if params != nil && params.Page != nil {
@@ -445,17 +444,23 @@ func processResponse(client *GroupMgmtClient, response *resty.Response, path str
 
 	} else {
 		// error response
-		return processError(response.StatusCode(), response.Body())
+		return nil,processError(response.StatusCode(), response.Body())
 	}
 }
 
-// append error code and error message
-func processError(httpCode int, body []byte) (interface{}, error) {
-	errResp, err := unwrapError(body)
+// process error response
+func processError(httpCode int, body []byte) (error) {
+	errResp := ""
+	wrapper := &ErrorResponse{}
+	err := json.Unmarshal(body, wrapper)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	return nil, fmt.Errorf("http response error: status (%d), messages: %v", httpCode, errResp)
+
+	for _, emsg := range wrapper.Messages {
+		errResp += fmt.Sprintf("%+v", *emsg)
+	}
+	return fmt.Errorf("error: http status(%d), messages: %v", httpCode, errResp)
 }
 
 //unwrap error response
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -88,11 +88,12 @@ func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool
 		AddRetryCondition(
 			func(resp *resty.Response, err error) bool {
 				if err == nil && resp.StatusCode() == 401 {
-
-					err := groupMgmtClient.refreshSessionToken()
+					// login again and refreh the session token
+					sessionToken, err = groupMgmtClient.login(username, password)
 					if err != nil {
 						return false
 					}
+					groupMgmtClient.SessionToken = sessionToken
 					resp.Request.SetHeader("X-Auth-Token", groupMgmtClient.SessionToken)
 					return true
 				}
@@ -105,28 +106,6 @@ func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool
 // EnableDebug : enables debug logging of client request/response
 func (client *GroupMgmtClient) EnableDebug() {
 	client.Client.SetDebug(true)
-}
-
-// refreshSessionToken : refresh session token
-func (client *GroupMgmtClient) refreshSessionToken() error {
-	// copy the client data to temp
-	tempClient := *client
-	// create a new client from copy
-	newClient := &tempClient
-
-	rclient := resty.New()
-	rclient.SetTLSClientConfig(&tls.Config{
-		InsecureSkipVerify: true,
-	})
-	newClient.Client = rclient
-	newSessionToken, err := newClient.login(client.Username, client.Password)
-	if err != nil {
-		return err
-	}
-	// set new auth token
-	client.SessionToken = newSessionToken
-	return nil
-
 }
 
 func (client *GroupMgmtClient) login(username, password string) (string, error) {
@@ -159,12 +138,7 @@ func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder
 	if err != nil {
 		return nil, err
 	}
-
-	// Unauthorize access, retry POST if Auth token is expired
-	if !response.IsSuccess() {
-		return nil, processError(response.StatusCode(), response.Body())
-	}
-	return processResponse(client, response, path, respHolder)
+	return processResponse(client, response, path, respHolder, nil)
 }
 
 // Put
@@ -182,11 +156,7 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
 	if err != nil {
 		return nil, err
 	}
-	// Unauthorize access, retry POST if Auth token is expired
-	if !response.IsSuccess() {
-		return nil, processError(response.StatusCode(), response.Body())
-	}
-	return processResponse(client, response, path, respHolder)
+	return processResponse(client, response, path, respHolder, nil)
 }
 
 // Get : Only used to get a single object with the given ID
@@ -206,12 +176,7 @@ func (client *GroupMgmtClient) Get(path string, id string, respHolder interface{
 	if response.StatusCode() == 404 {
 		return nil, nil
 	}
-
-	// Unauthorize access, retry POST if Auth token is expired
-	if !response.IsSuccess() {
-		return nil, processError(response.StatusCode(), response.Body())
-	}
-	return processResponse(client, response, path, respHolder)
+	return processResponse(client, response, path, respHolder, nil)
 }
 
 // Delete :
@@ -225,13 +190,7 @@ func (client *GroupMgmtClient) Delete(path string, id string) error {
 	if err != nil {
 		return err
 	}
-
-	// Unauthorize access, retry Delete if Auth token is expired
-	// Unauthorize access, retry POST if Auth token is expired
-	if !response.IsSuccess() {
-		return processError(response.StatusCode(), response.Body())
-	}
-	_, err = processResponse(client, response, path, nil)
+	_, err = processResponse(client, response, path, nil, nil)
 	return err
 }
 
@@ -318,20 +277,7 @@ func (client *GroupMgmtClient) listPost(
 	if err != nil {
 		return nil, err
 	}
-	// Unauthorize access, retry listPost if Auth token is expired
-	// Unauthorize access, retry POST if Auth token is expired
-	if !response.IsSuccess() {
-		return nil, processError(response.StatusCode(), response.Body())
-	}
-
-	if params != nil && params.Page != nil {
-		totalRows, err := getTotalRows(response.Body())
-		if err != nil {
-			return nil, err
-		}
-		params.Page.TotalRows = totalRows
-	}
-	return processResponse(client, response, path, nil)
+	return processResponse(client, response, path, nil, params)
 }
 
 // listGet uses a get request to get all objects on the path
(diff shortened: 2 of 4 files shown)


pkg/client/group_mgmt_client.go
```diff
@@ -84,6 +88,21 @@ func (client *GroupMgmtClient) EnableDebug() {
 	client.Client.SetDebug(true)
 }
 
+// refreshSessionToken : refresh session token
+func (client *GroupMgmtClient) refreshSessionToken(code int, message string) (bool, error) {
+	if code >= 400 {
```
rkumpf: Is there a reason to check for anything other than 401?

pkg/client/group_mgmt_client.go
```diff
@@ -84,6 +88,21 @@ func (client *GroupMgmtClient) EnableDebug() {
...
+			newSessionToken, err := client.login(client.Username, client.Password)
+			if err != nil {
+				return false, err
+			}
+			client.SessionToken = newSessionToken
+			return true, nil
+		}
+	}
+	return false, nil
+}
```
rkumpf: A blank line would be nice here.

pkg/client/group_mgmt_client.go
```diff
@@ -113,12 +133,22 @@ func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder
...
 	// IsSuccess method returns true if HTTP status `code >= 200 and <= 299` otherwise false.
 	if !response.IsSuccess() {
 		errResp, err := unwrapError(response.Body())
 		if err != nil {
 			return nil, err
 		}
+		// refresh session token
+		isSessionRefresh, err := client.refreshSessionToken(response.StatusCode(), errResp)
+		if err != nil {
+			return nil, err
```
rkumpf: It might be unexpected for the caller to receive an error that did not come in response to their request.  An alternative is to return the original 401 error, but that doesn't seem better.

Will we advertise the automatic session management somewhere, like in the README?  That might be the place to mention this little quirk.

README.md
```diff
@@ -1 +1,22 @@
-# nimble-golang-sdk
\ No newline at end of file
+# HPE Nimble Storage SDK in Go Programming Language
+This is a Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
+
+The SDK provides a Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
```
rkumpf: Remove the word "a" in the first sentence?

It is worth mentioning get/read operations in the third sentence.

pkg/service/volume_service_test.go
```diff
@@ -318,6 +324,37 @@ func (suite *VolumeServiceTestSuite) TestACLVolume() {
 	}
 }
 
+func (suite *VolumeServiceTestSuite) TestExpiredToken() {
+
+	volume := suite.createVolume("TestExpiredToken")
+	if volume != nil {
+		// Get the token expiry time
+		sToken := suite.groupService.client.SessionToken
```
rkumpf: Could we do something simple like this to avoid the long wait?

suite.groupService.client.SessionToken = "expired"

pkg/client/group_mgmt_client.go
```diff
@@ -84,6 +88,20 @@ func (client *GroupMgmtClient) EnableDebug() {
 	client.Client.SetDebug(true)
 }
 
+// refreshSessionToken : refresh session token
+func (client *GroupMgmtClient) refreshSessionToken(code int, message string) (bool, error) {
+	if code >= 400 && client.SessionToken != "" {
+		if strings.Contains(message, "SM_http_unauthorized") {
+			newSessionToken, err := client.login(client.Username, client.Password)
```
rkumpf: Before this line, add:

client.SessionToken = ""

Otherwise a changed password can still cause infinite recursion.

pkg/service/volume_service_test.go
vidhutsingh: Yes we can do that . thanks 

`func (suite *VolumeServiceTestSuite) TestExpiredToken() {

	volume := suite.createVolume("TestExpiredToken")
	if volume != nil {
		//Set invalid session Token 9d255b36c700ec8b56e2064e67f01c45
		suite.groupService.client.SessionToken = "9d255b36c700ec8b56e2064e67f01c45"
		suite.deleteVolume("TestExpiredToken")
	}
}`

README.md
vidhutsingh: I will remove "a" . Please note that this is the initial README. It will be rewritten by Michael.

pkg/client/group_mgmt_client.go
vidhutsingh: Just want to avoid any NOS version issue where it doesn't return 401 for unauthorized error . Ideally we should use 401 .

pkg/client/group_mgmt_client.go
vidhutsingh: Done .

pkg/client/group_mgmt_client.go
vidhutsingh: May contain in the documentation.

pkg/client/group_mgmt_client.go
vidhutsingh: done

README.md
```diff
@@ -1 +1,22 @@
-# nimble-golang-sdk
\ No newline at end of file
+# HPE Nimble Storage SDK in Go Programming Language
+This is Go Software Development Kit (SDK) for [HPE Nimble Storage](http://hpe.com/storage/nimblestorage) arrays. The HPE Nimble Storage array has a Representational State Transfer (REST) web service application programming interface (API). The SDK implements a simple client Go modules for communicating with the HPE Nimble Storage REST API. The Go module is being used to communicate with the API over HTTPS.
```
rgcostea: This is *the* Go Software...

README.md
```diff
@@ -1 +1,22 @@
...
+
+The SDK provides a Go modules to interact with the HPE Nimble Storage REST API. The code abstracts the lower-level API calls into Go objects that you can easily incorporate into any automation or DevOps workflows. Use it to create, modify and delete most resources like volumes, volume collections, initiator groups and more, as well as perform other tasks like snapshotting, cloning, restoring data, etc.
```
rgcostea: The SDK provides Go modules...

pkg/client/group_mgmt_client.go
rgcostea: We should only do this for a 401.  All other 400 errors need to make it back to the client for assessment.

400 (bad request) - could be a malformed URL or post body
403 (forbidden) - permission issue or inability to perform an operation on some object
404 (not found) - bad url path
409 (conflict) - object in use
...

vidhutsingh: @rgcostea , we are returning all the error code to the client. Here we are checking both code >= 400 and SM_http_unauthorized before considering this as an expired token scenario. So if you are certain that we don't get any other error code with any previous NOS version for unauthorized error, I will make this change.

pkg/client/group_mgmt_client.go
```diff
@@ -159,6 +190,15 @@ func (client *GroupMgmtClient) Put(path, id string, payload interface{}, respHol
...
 		}
+		// refresh session token
+		isSessionRefresh, err := client.refreshSessionToken(response.StatusCode(), errResp)
+		if err != nil {
+			return nil, err
+		}
+		if isSessionRefresh {
+			// retry the ops
+			goto RETRY
+		}
```
rgcostea: goto statements are rarely a good option when you can opt to do something else.

Why is this code copied in every function (Get, Post, Put, etc.)?  Why not simply move this code into a function (e.g. verifyAuth) which checks for the 401 code and attempts to recreate the session?

Regarding the goto, please investigate the RetryCondition option in go-resty.  If that doesn't pan out, I still prefer we find another way to perform the retry rather than a goto.

pkg/service/group_service_factory.go
```diff
@@ -72,6 +74,26 @@ func NewNsGroupService(ip, username, password, apiVersion string, synchronous bo
...
+			Value:     sessionToken,
+		},
+	}
+	tokenList, err := tokenService.GetTokens(sFilter)
+	if err != nil {
+		return err
+	}
+	return tokenService.DeleteToken(*tokenList[0].ID)
+}
+
```
rgcostea: This doesn't seem correct to me.  I prefer you just delete the token through the tokenService.  Alternatively, you can rename this function to logout I suppose.

pkg/client/group_mgmt_client.go
rgcostea: I'm certain.  It's only ever a 401.  Let's keep it clean.

pkg/client/group_mgmt_client.go
vidhutsingh: >> Why is this code copied in every function (Get, Post, Put, etc.)? Why not simply move this code into a function (e.g. verifyAuth) which checks for the 401 code and attempts to recreate the session?

We need refresh the token in case of 401 and also we need to retry the failed operation so it has to go in (Get, Post, Put etc ). 

I will check go-resty , other way could be to use for{} loop instead of goto.

suneeth51 (dismissed):
@vidhutsingh  As @rgcostea  mentioned, definitely we can further optimize the token refresh logic and also refactor the code across all those CRUD operations and have **single common function** that first attempts perform REST request to the Array and if it encounters **401** error, then only you need to perform token refreshing.

Please refer to this CSP implementation for reference:
https://github.com/hpe-storage/common-host-libs/blob/master/storageprovider/csp/container_storage_provider.go
*func (provider *ContainerStorageProvider) invoke(request *connectivity.Request) (status int, err error)*

vidhutsingh:
Thanks @suneeth51

vidhutsingh:
> https://github.com/hpe-storage/common-host-libs/blob/master/storageprovider/csp/container_storage_provider.go

We shouldn't compare this with CSP implementation as here we are using resty package and each CRUD operations GET, PUT etc are implemented as different function unlike in CSP where you build the path and just invoke "provider.Client.DoJSON()"
Here is code for reference "
`response, err := client.Client.R().
		SetHeader("X-Auth-Token", client.SessionToken).
		SetBody(&DataWrapper{
			Data: payload,
		}).
		Post(url)`

I will still see how much refactoring is possible although sure we can remove GOTO and use recursion call instead.

rgcostea:
I feel the RetryCondition is a better option.  You can perform a login, update the header token, and return true.  This will force the resty client to retry.

pkg/service/group_service_factory.go
vidhutsingh: Ok will change to LogoutService

README.md
vidhutsingh: ok thanks

README.md
vidhutsingh: sure

pkg/client/group_mgmt_client.go
```diff
@@ -84,6 +88,23 @@ func (client *GroupMgmtClient) EnableDebug() {
 	client.Client.SetDebug(true)
 }
 
+// refreshSessionToken : refresh session token
+func (client *GroupMgmtClient) refreshSessionToken() (bool, error) {
+	// Invalid credential has empty token
+	if client.SessionToken != "" {
```
rkumpf: I've lost track of what this condition is for.  I'm concerned that if the login fails at line 96, we cannot recover.  Is there any update on using RetryConditions?

vidhutsingh:
> I feel the RetryCondition is a better option. You can perform a login, update the header token, and return true. This will force the resty client to retry.

Yes @rgcostea , resty RetryCondition is a superior option although it took me a while to understand how does it work, it is totally worth it. 
thanks . 👍

vidhutsingh:
> > I feel the RetryCondition is a better option. You can perform a login, update the header token, and return true. This will force the resty client to retry.
> 
> Yes @rgcostea , resty RetryCondition is a superior option although it took me a while to understand how does it work, it is totally worth it.
> thanks . 👍

While doing unit testing discover an issue, the RetryCondition is not working in our case where Auth Token is refreshed. It always pic the expired token. Although workflow retry does happen possibly a bug or by design with rest since 
the request doesn't take a pointer to SessionToken it takes string value. 

```
response, err := client.Client.R().
		SetHeader("X-Auth-Token", client.SessionToken).
		SetBody(&DataWrapper{
			Data: payload,
		}).
```

rgcostea:
What does your code look like in the RetryCondition?

vidhutsingh:
```
Here is the code with only changes 
``// NewClient instantiates a new client to communicate with the Nimble group
func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
	// Create GroupMgmt Client
	client := resty.New()
	client.SetTLSClientConfig(&tls.Config{
		InsecureSkipVerify: true,
	})
	groupMgmtClient := &GroupMgmtClient{
		URL:       fmt.Sprintf("https://%s:5392/%s", ipAddress, apiVersion),
		Client:    client,
		WaitOnJob: waitOnJobs,
		Username:  username,
		Password:  password,
	}

	// Get session token
	sessionToken, err := groupMgmtClient.login(username, password)
	if err != nil {
		return nil, err
	}
	groupMgmtClient.SessionToken = sessionToken

	// Add retryCondition
	groupMgmtClient.Client.
		SetRetryCount(maxOpsRetries).
		AddRetryCondition(
			func(resp *resty.Response, err error) bool {
				if err == nil && resp.StatusCode() == 401 {

					err := groupMgmtClient.refreshSessionToken()
					if err != nil {
						return false
					}
					fmt.Println("RETRY WORKFLOW with new token", groupMgmtClient.SessionToken)
					return true
				}
				return false
			},
		)
	return groupMgmtClient, nil
}

// EnableDebug : enables debug logging of client request/response
func (client *GroupMgmtClient) EnableDebug() {
	client.Client.SetDebug(true)
}

// refreshSessionToken : refresh session token
func (client *GroupMgmtClient) refreshSessionToken() error {
	nClient := *client
	newClient := &nClient
	rclient := resty.New()
	rclient.SetTLSClientConfig(&tls.Config{
		InsecureSkipVerify: true,
	})
	newClient.Client = rclient
	newSessionToken, err := newClient.login(client.Username, client.Password)
	if err != nil {
		return err
	}
	// set new auth token
	fmt.Println("new session", newSessionToken)
	client.SessionToken = newSessionToken
	return nil

}``
```

rgcostea:
It looks like you are setting the retrycondition on the Client and not on the request:

groupMgmtClient.Client vs client.Client.R() 

client.Client.R() is where the auth-token is set.

vidhutsingh:
As per the below example code it has to be set on client 
```
retryCount := 5
	retryIntervals := make([]uint64, retryCount+1)

	// Set retry wait times that do not intersect with default ones
	retryWaitTime := time.Duration(3) * time.Second
	retryMaxWaitTime := time.Duration(9) * time.Second

	c := dc().
		SetRetryCount(retryCount).
		SetRetryWaitTime(retryWaitTime).
		SetRetryMaxWaitTime(retryMaxWaitTime).
		AddRetryCondition(
			func(r *Response, _ error) bool {
				timeSlept, _ := strconv.ParseUint(string(r.Body()), 10, 64)
				retryIntervals[attempt] = timeSlept
				attempt++
				return true
			},
		)
	_, _ = c.R().Get(ts.URL + "/set-retrywaittime-test")
```

Reference : https://github.com/go-resty/resty/blob/master/resty_test.go

rgcostea:
Ok.  But the Response in the RetryCondition has a reference to the request.  Can't you call 

`response.request.SetHeader("X-Auth-Token", newSession)`

vidhutsingh:
It worked.  thanks. 
How did you find out that response can also be used to set request header? that was really nice, seems quirky a bit :) 

Please check the latest code

rgcostea:
I just read the docs: https://godoc.org/github.com/go-resty/resty#Response

vidhutsingh:
Thanks @rgcostea . I just realised that we are also storing the rest client pointer in the groupMgtClient, we could have used it to set the header as well. Either way is fine.

pkg/client/group_mgmt_client.go
```diff
@@ -76,14 +81,54 @@ func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool
...
+// refreshSessionToken : refresh session token
+func (client *GroupMgmtClient) refreshSessionToken() error {
+	// copy the client data to temp
+	tempClient := *client
+	// create a new client from copy
+	newClient := &tempClient
+
+	rclient := resty.New()
+	rclient.SetTLSClientConfig(&tls.Config{
+		InsecureSkipVerify: true,
```
rkumpf: We should clearly document this choice.  It's probably not a first release requirement to do something better, but at some point we should.  I hope George's request to change the constructor to use the functional options pattern will be implemented, since it will let us make these kinds of changes without breaking API compatibility.

pkg/client/group_mgmt_client.go
```diff
@@ -76,14 +81,54 @@ func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool
...
+// EnableDebug : enables debug logging of client request/response
 func (client *GroupMgmtClient) EnableDebug() {
 	client.Client.SetDebug(true)
 }
 
+// refreshSessionToken : refresh session token
+func (client *GroupMgmtClient) refreshSessionToken() error {
+	// copy the client data to temp
+	tempClient := *client
+	// create a new client from copy
```
rkumpf: Let's add a comment explaining why we create a new client.  I guess it is because the resty client gets upset if it has two requests going in parallel.

pkg/client/group_mgmt_client.go
```diff
@@ -113,69 +159,34 @@ func (client *GroupMgmtClient) Post(path string, payload interface{}, respHolder
 	if err != nil {
 		return nil, err
 	}
-	// IsSuccess method returns true if HTTP status `code >= 200 and <= 299` otherwise false.
+
+	// Unauthorize access, retry POST if Auth token is expired
```
rkumpf: This isn't true anymore, right?  I think the retry would already have happened at line 158.

pkg/client/group_mgmt_client.go
```diff
@@ -359,30 +351,36 @@ func (client *GroupMgmtClient) listGet(
...
 	}
+	return processResponse(client, response, path, nil)
+}
 
+// unwrapData
+func getTotalRows(body []byte) (*int, error) {
 	// unmarshal the response
 	wrapper := &DataWrapper{}
-	err = json.Unmarshal(response.Body(), wrapper)
+	err := json.Unmarshal(body, wrapper)
```
rkumpf: It's a shame we have to unmarshal the response twice.  Could we reasonably avoid it?

pkg/client/group_mgmt_client.go
```diff
@@ -391,11 +389,53 @@ func unwrap(body []byte, payload interface{}) (interface{}, error) {
...
-
 	// return it
 	return wrapper.Data, nil
 }
 
+//processResponse
+func processResponse(client *GroupMgmtClient, response *resty.Response, path string, respHolder interface{}) (interface{}, error) {
+
+	// process successfull response
+	if response.IsSuccess() {
```
rkumpf, line 320: In several places I see logic like this:

        if !response.IsSuccess() {
                return nil, processError(response.StatusCode(), response.Body())
        }
        return processResponse(client, response, path, respHolder)

It seems redundant with this check.

pkg/service/volume_service_test.go
```diff
@@ -23,18 +23,23 @@ type VolumeServiceTestSuite struct {
...
 func (suite *VolumeServiceTestSuite) config() *NsGroupService {
+
 	groupService, err := NewNsGroupService("10.18.174.8", "admin", "admin", "v1", true)
 	if err != nil {
 		suite.T().Errorf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
 		return nil
 	}
 	// set debug
-	//groupService.SetDebug()
+	groupService.SetDebug()
```
rkumpf: Is this change desired?

pkg/service/volume_service_test.go
```diff
@@ -112,6 +122,9 @@ func (suite *VolumeServiceTestSuite) createVolColl(name string) {
 }
 
 func (suite *VolumeServiceTestSuite) TestGetNonExistentVolumeByID() {
+	if suite.groupService == nil {
```
rkumpf: What a bummer to put these checks everywhere.  Is there no better option?

rkumpf (commented):
I like the new RetryCondition.  :)

pkg/client/group_mgmt_client.go
rgcostea: Why do we create a new client?  I don't understand the need.

pkg/client/group_mgmt_client.go
rgcostea: +1

I think the tls.Config should be an option to the GroupMgmtClientConfig.  Let the user decide if they want to skip trust to the array.

pkg/client/group_mgmt_client.go
```diff
@@ -223,15 +226,13 @@ func (client *GroupMgmtClient) Delete(path string, id string) error {
 		return err
 	}
 
-	// http Error
+	// Unauthorize access, retry Delete if Auth token is expired
+	// Unauthorize access, retry POST if Auth token is expired
```
rgcostea: Delete or Post?

pkg/service/volume_service_test.go
rgcostea: Agreed.  Why were they added?

It seems since you check for this SetupTest() above, it will fail there and none of these tests will run.

pkg/client/group_mgmt_client.go
vidhutsingh: Yes it is not needed now since we are setting the header . Infact whole refreshSessionToken fn is not required .

pkg/client/group_mgmt_client.go
vidhutsingh: Leftover comments , will clean up

pkg/client/group_mgmt_client.go
vidhutsingh: Ok will fix it .

pkg/client/group_mgmt_client.go
vidhutsingh: Oh we dont need it , let me clean up

pkg/service/volume_service_test.go
vidhutsingh: No

pkg/service/volume_service_test.go
vidhutsingh: It runs and panic in all the test cases . Thus had to put nil check everywhere.

pkg/client/group_mgmt_client.go
vidhutsingh: The comment is not needed here .

pkg/service/volume_service_test.go
vidhutsingh: No it doesnt , it runs all the test even setup failed.

pkg/client/group_mgmt_client.go
vidhutsingh: done

pkg/client/group_mgmt_client.go
vidhutsingh: yes

pkg/client/group_mgmt_client.go
vidhutsingh: not needed . will cleanup .

pkg/client/group_mgmt_client.go
vidhutsingh: We will track it via functional option pattern

pkg/service/volume_service_test.go
rkumpf: If setup fails, with this change does the test case pass or fail?

pkg/client/group_mgmt_client.go
```diff
@@ -411,9 +330,12 @@ func processResponse(client *GroupMgmtClient, response *resty.Response, path str
 				return nil, err
 			}
 		}
-		// process success response
-
-		return unwrapData(response.Body(), respHolder)
+		// unmarshall response data
+		dataIntf, totalRows, err := unwrapData(response.Body(), respHolder)
+		if params != nil && params.Page != nil && totalRows != nil {
```
rkumpf, line 335: I guess the 'totalRows != nil' check doesn't really serve a purpose.

pkg/client/group_mgmt_client.go
```diff
@@ -411,9 +330,12 @@ func processResponse(client *GroupMgmtClient, response *resty.Response, path str
...
+			params.Page.TotalRows = totalRows
```
rkumpf, line 336: This comment isn't related to this code review, but it is a convenient place for me to mention it anyway.  :)

The pagination.go file is in a param/pagination package.  Is there a reason not to put it in the param package directly?

pkg/service/volume_service_test.go
vidhutsingh: setup is not a test , all the test starts with Test* methods so eventually it will come to all the Test* methods even after setup has error .

vidhutsingh: Let me check the documentation of suite package . will see if anything can be done about it .

pkg/client/group_mgmt_client.go
```diff
@@ -358,42 +296,66 @@ func (client *GroupMgmtClient) listGet(
...
+		// http code 202, handle async job
+		if response.StatusCode() == 202 {
+			id, err := processAsyncResponse(client, response.Body())
+			if id != nil {
+				// action rpc may contain different path.
+				// extract Get uri path from original path.
+				newPath := strings.Split(path, "/")
+				return client.Get(newPath[0], id.(string), respHolder)
+			} else {
+				return nil, err
```
rkumpf, line 330: Nit: It might be more readable if we remove the 'else' and unindent this 'return'.

pkg/service/volume_service_test.go
vidhutsingh: Found it , we just need to exit the suite. However there is no way via suite package . pls check last commit

pkg/client/group_mgmt_client.go
```diff
@@ -358,42 +296,66 @@ func (client *GroupMgmtClient) listGet(
...
+			}
+		}
+		// unmarshall response data
```
rkumpf, line 333: The status code is presumed to be 200 OK when we get here, right?  I suggest checking for that and returning some kind of "unexpected status code" error otherwise.

pkg/client/group_mgmt_client.go
```diff
@@ -358,42 +296,66 @@ func (client *GroupMgmtClient) listGet(
...
 	err := json.Unmarshal(body, wrapper)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	// return it
-	return wrapper.Data, nil
+	for _, emsg := range wrapper.Messages {
+		errResp += fmt.Sprintf("%+v", *emsg)
```
rkumpf, line 356: We need a delimiter between the messages.

rkumpf (commented):
Please consider addressing these mostly minor and/or unrelated comments in the next patch.

pkg/service/volume_service_test.go
```diff
@@ -24,7 +25,7 @@ type VolumeServiceTestSuite struct {
 
 func (suite *VolumeServiceTestSuite) config() *NsGroupService {
 
-	groupService, err := NewNsGroupService("10.18.174.8", "admin", "admin", "v1", true)
+	groupService, err := NewNsGroupService("10.18.174.8", "admin", "admina", "v1", true)
```
rkumpf: Do not include a password.  One option is to require an environment variable to be set when the test is run.

rkumpf (approved):

pkg/client/group_mgmt_client.go
vidhutsingh: We need to initialze the totalRow which is not visible outside this pkg

pkg/client/group_mgmt_client.go
vidhutsingh: Not require since error is already captured in err

pkg/service/volume_service_test.go
vidhutsingh: That is a test code . that should be okay .

pkg/client/group_mgmt_client.go
vidhutsingh: It does since if there is error it will be nil .

vidhutsingh:
I think all the major changes have been accommodated with the last commit.

pkg/service/volume_service_test.go
rkumpf: It is not okay.  Email me if you would like to discuss it.

pkg/client/group_mgmt_client.go
rkumpf: I'm referring to the handling of status codes like 201 Created, 204 No Content, etc.

pkg/client/group_mgmt_client.go
rkumpf: I probably wasn't clear.  I'm suggesting moving param/pagination/pagination.go to param/pagination.go, unless there is a good reason to have a separate package for it.

pkg/client/group_mgmt_client.go
rkumpf: I'm saying this would be enough:

		if params != nil && params.Page != nil {
			params.Page.TotalRows = totalRows

If totalRows is nil, I would think params.Page.TotalRows should be nil also.

rkumpf: I'm reconsidering this one.  I think what you mean is when an error occurs, we should preserve the prior TotalRows value even though we'll be returning something that didn't come back in the API response.  The client program will see that an error is returned and know not to inspect the TotalRows value.  That seems reasonable to me.  We can close this issue.

pkg/service/volume_service_test.go
vidhutsingh: removed it

pkg/service/volume_service_test.go
rkumpf: It's back! ;)

rkumpf (approved):

rkumpf (approved):

vidhutsingh:
@rkumpf, the merge is still not approved.

vidhutsingh:
@rgcostea , there is one approver but still, the merge is not allowed. is there any setting change?

rkumpf:
The merge is blocked because Suneeth has an outstanding request for changes.  His comment seems to be addressed acceptably in the updated patch, but I don't perceive sufficient urgency to warrant the bold step of dismissing his review.

vidhutsingh:
> The merge is blocked because Suneeth has an outstanding request for changes. His comment seems to be addressed acceptably in the updated patch, but I don't perceive sufficient urgency to warrant the bold step of dismissing his review.

Nobody suggested that . It just that I wasn't aware of what is blocking the merge. now it's clear. ;)

rkumpf:
I know you didn't suggest it, but I was considering it nonetheless....

vidhutsingh:
@suneeth51 , the merge is blocked because of your review change request. If you don't have any further comments, please approve. 

Very poor feature of GitHub as any reviewer can block the merge. Bummer!

vidhutsingh:
I am blocked due to this .

vidhutsingh:
@rkumpf , thank you!

suneeth51:
@rkumpf Thanks for taking care of this to unblock @vidhutsingh. The changes looks good.

PR #14, opened 2020-10-15
con-1497: Add interface to groupfactory to help with unit testing

ckallianpur:
Add interface to group_service to help with unit testing in the array-exporter

### DIFF
--- a/pkg/service/group_service_factory.go
+++ b/pkg/service/group_service_factory.go
@@ -4,10 +4,17 @@ package service
 
 import (
 	"fmt"
+
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 )
+
+//GroupServiceFactory - Interface for the group_service_factory
+type GroupServiceFactory interface {
+	GetVolumeService() (s VolumeServicer)
+	GetPoolService() (s PoolServicer)
+}
 
 // NsGroupService type
 type NsGroupService struct {
@@ -21,7 +28,7 @@ type NsGroupService struct {
 	masterKeyService                  *MasterKeyService
 	alarmService                      *AlarmService
 	subscriptionService               *SubscriptionService
-	volumeService                     *VolumeService
+	volumeService                     VolumeServicer
 	shelfService                      *ShelfService
 	keyManagerService                 *KeyManagerService
 	protectionTemplateService         *ProtectionTemplateService
@@ -54,7 +61,7 @@ type NsGroupService struct {
 	witnessService                    *WitnessService
 	jobService                        *JobService
 	auditLogService                   *AuditLogService
-	poolService                       *PoolService
+	poolService                       PoolServicer
 	volumeCollectionService           *VolumeCollectionService
 	subscriberService                 *SubscriberService
 	diskService                       *DiskService
@@ -151,7 +158,7 @@ func (gs *NsGroupService) GetSubscriptionService() (vs *SubscriptionService) {
 }
 
 // GetVolumeService - returns service of a type *VolumeService
-func (gs *NsGroupService) GetVolumeService() (vs *VolumeService) {
+func (gs *NsGroupService) GetVolumeService() (vs VolumeServicer) {
 	if gs.volumeService == nil {
 		gs.volumeService = NewVolumeService(gs)
 	}
@@ -415,7 +422,7 @@ func (gs *NsGroupService) GetAuditLogService() (vs *AuditLogService) {
 }
 
 // GetPoolService - returns service of a type *PoolService
-func (gs *NsGroupService) GetPoolService() (vs *PoolService) {
+func (gs *NsGroupService) GetPoolService() (vs PoolServicer) {
 	if gs.poolService == nil {
 		gs.poolService = NewPoolService(gs)
 	}
--- a/pkg/service/volume_service.go
+++ b/pkg/service/volume_service.go
@@ -7,10 +7,31 @@ package service
 
 import (
 	"fmt"
+
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 )
+
+type VolumeServicer interface {
+	GetVolumes(params *param.GetParams) ([]*nimbleos.Volume, error)
+	CreateVolume(obj *nimbleos.Volume) (*nimbleos.Volume, error)
+	UpdateVolume(id string, obj *nimbleos.Volume) (*nimbleos.Volume, error)
+	GetVolumeByID(id string) (*nimbleos.Volume, error)
+	GetVolumeByName(name string) (*nimbleos.Volume, error)
+	GetVolumeBySerialNumber(serialNumber string) (*nimbleos.Volume, error)
+	OnlineVolume(id string, force bool) (*nimbleos.Volume, error)
+	OfflineVolume(id string, force bool) (*nimbleos.Volume, error)
+	DeleteVolume(id string) error
+	AssociateVolume(volID string, volcollID string) error
+	DisassociateVolume(volID string) error
+	RestoreVolume(id string, baseSnapID string) error
+	MoveVolume(id string, destPoolID string, forceVvol *bool) (*nimbleos.NsVolumeListReturn, error)
+	BulkMoveVolumes(volIds []*string, destPoolID string, forceVvol *bool) (*nimbleos.NsVolumeListReturn, error)
+	AbortMoveVolume(id string) error
+	BulkSetDedupeVolumes(volIds []*string, dedupeEnabled bool) error
+	BulkSetOnlineAndOfflineVolumes(volIds []*string, online bool) error
+}
 
 // VolumeService type
 type VolumeService struct {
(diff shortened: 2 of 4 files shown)


pkg/service/group_service_factory.go
```diff
@@ -20,7 +22,7 @@ type NsGroupService struct {
 	chapUserService                   *ChapUserService
 	masterKeyService                  *MasterKeyService
 	alarmService                      *AlarmService
-	volumeService                     *VolumeService
+	volumeService                     servicefactory.VolumeServiceHelper
```
vidhutsingh: This change will break the SDK since you are replacing volumService with servicefactory.VolumeServiceHelper .

vidhutsingh: This change will break the SDK since you are replacing volumService with servicefactory.VolumeServiceHelper. You better add your new helper pointer. Same comment is for other changes .

ckallianpur: This is just a sample. Was thinking of creating an interface for all services and using that here.  Is there an automated way to do that?

rkumpf (commented):
First we need to decide whether to accept this testing support into the SDK.  From what I can tell, it does seem like the Golang thing to do.  Assuming we do...

I think the service interfaces belong directly in the service classes.  For example, VolumeService (perhaps renamed) would be the interface, and NewVolumeService() would return an instance of the "real" implementation of it.  Then we wouldn't need the *ServiceHelper interfaces, which I find confusing.

We would want full-blown "fake" versions of all the services.  I.e., a service/fake package full of mock service files.  I don't yet fully understand the idiom of how their behavior is controlled by the unit test, but presumably the values they return would be pretty bland.  Perhaps uninitialized values (not nil pointers) for every field.  Ideally the mock service would define the object(s) to be returned in such a way that unit tests could replace their contents.

rgcostea:
> First we need to decide whether to accept this testing support into the SDK. From what I can tell, it does seem like the Golang thing to do. Assuming we do...
> 
> I think the service interfaces belong directly in the service classes. For example, VolumeService (perhaps renamed) would be the interface, and NewVolumeService() would return an instance of the "real" implementation of it. Then we wouldn't need the *ServiceHelper interfaces, which I find confusing.
> 
> We would want full-blown "fake" versions of all the services. I.e., a service/fake package full of mock service files. I don't yet fully understand the idiom of how their behavior is controlled by the unit test, but presumably the values they return would be pretty bland. Perhaps uninitialized values (not nil pointers) for every field. Ideally the mock service would define the object(s) to be returned in such a way that unit tests could replace their contents.

Agreed.  We shouldn't use a ServiceHelper.  Using a fake and a real implementation of an interface type is the way to go.  An example of how we did this elsewhere can be found here: https://github.com/hpe-storage/common-host-libs/tree/master/storageprovider

pkg/service/fake/fake_pool_service.go
```diff
@@ -0,0 +1,89 @@
+package fake
```
rgcostea: missing copyright

pkg/service/fake/fake_volume_service.go
```diff
@@ -0,0 +1,148 @@
+package fake
```
rgcostea: copyright

pkg/service/volume_service.go
```diff
@@ -7,11 +7,32 @@ package service
 
 import (
 	"fmt"
+
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 )
 
+type IVolumeService interface {
```
rgcostea: I personally don't like this naming convention.  I'd much rather see a VolumeService interface and then a Fake and a Nimble implementation.  The client doesn't know at the end of the day as we would encapsulate all of that.

pkg/service/group_service_factory.go
```diff
@@ -53,7 +55,7 @@ type NsGroupService struct {
 	witnessService                    *WitnessService
 	jobService                        *JobService
 	auditLogService                   *AuditLogService
-	poolService                       *PoolService
+	poolService                       servicefactory.PoolServiceHelper
```
rkumpf: I think Vidhut is saying these need to be pointers.

ckallianpur: I was getting an error doing that. Since IVolumeService is a interface which contains a pointer to struct, we can use it directly? https://stackoverflow.com/questions/44370277/type-is-pointer-to-interface-not-interface-confusion

rgcostea (changes requested):
Please address review comments

vidhutsingh:
Overall this approach will unit test only service layer which is only part of the sdk code. Ideally we should test service layer, object set layer and added sdk features like pagination etc. If there is a plan to fake it , shouldn't we fake the array  ? 
One of the approch could be to have fake_client implementation so that unit test goes through all the layers of the sdk.

rkumpf:
> Overall this approach will unit test only service layer which is only part of the sdk code. Ideally we should test service layer, object set layer and added sdk features like pagination etc. If there is a plan to fake it , shouldn't we fake the array ?
> One of the approch could be to have fake_client implementation so that unit test goes through all the layers of the sdk.

Perhaps the purpose is not clear.  I see two objectives:

1. Enable unit testing of the SDK (without connecting to an array).
2. Enable unit testing of a program that uses the SDK.

This patch targets objective #2.  Specifically, the ContainerVision project needs a "fake" version of the SDK with predictable behavior and no array dependency.

For objective #1, I agree that faking the RESTy client layer (or equivalent) is the thing to do.

ckallianpur:
> > Overall this approach will unit test only service layer which is only part of the sdk code. Ideally we should test service layer, object set layer and added sdk features like pagination etc. If there is a plan to fake it , shouldn't we fake the array ?
> > One of the approch could be to have fake_client implementation so that unit test goes through all the layers of the sdk.
> 
> Perhaps the purpose is not clear. I see two objectives:
> 
> 1. Enable unit testing of the SDK (without connecting to an array).
> 2. Enable unit testing of a program that uses the SDK.
> 
> This patch targets objective #2. Specifically, the ContainerVision project needs a "fake" version of the SDK with predictable behavior and no array dependency.
> 
> For objective #1, I agree that faking the RESTy client layer (or equivalent) is the thing to do.

Agree with @rkumpf. Here is the first pass on unit test using the fake package https://github.hpe.com/nimble-dcs/array-exporter/pull/29

pkg/service/volume_service.go
vidhutsingh: Agree, the naming convention is a bit off.

ckallianpur: Created https://nimblejira.nimblestorage.com/browse/SA-172 to run the generator and create all these fake services and interfaces. Should this PR be closed if SA-172 is going to take care of this?

vidhutsingh:
> > Overall this approach will unit test only service layer which is only part of the sdk code. Ideally we should test service layer, object set layer and added sdk features like pagination etc. If there is a plan to fake it , shouldn't we fake the array ?
> > One of the approch could be to have fake_client implementation so that unit test goes through all the layers of the sdk.
> 
> Perhaps the purpose is not clear. I see two objectives:
> 
> 1. Enable unit testing of the SDK (without connecting to an array).
> 2. Enable unit testing of a program that uses the SDK.
> 
> This patch targets objective #2. Specifically, the ContainerVision project needs a "fake" version of the SDK with predictable behavior and no array dependency.
> 
> For objective #1, I agree that faking the RESTy client layer (or equivalent) is the thing to do.

I was under impression that this is for SDK unit testing. Why would we even want to have SDK user unit testing code to be implemented here in this repo . This doesn't look right to me.

ckallianpur:
> > First we need to decide whether to accept this testing support into the SDK. From what I can tell, it does seem like the Golang thing to do. Assuming we do...
> > I think the service interfaces belong directly in the service classes. For example, VolumeService (perhaps renamed) would be the interface, and NewVolumeService() would return an instance of the "real" implementation of it. Then we wouldn't need the *ServiceHelper interfaces, which I find confusing.
> > We would want full-blown "fake" versions of all the services. I.e., a service/fake package full of mock service files. I don't yet fully understand the idiom of how their behavior is controlled by the unit test, but presumably the values they return would be pretty bland. Perhaps uninitialized values (not nil pointers) for every field. Ideally the mock service would define the object(s) to be returned in such a way that unit tests could replace their contents.
> 
> Agreed. We shouldn't use a ServiceHelper. Using a fake and a real implementation of an interface type is the way to go. An example of how we did this elsewhere can be found here: https://github.com/hpe-storage/common-host-libs/tree/master/storageprovider

ckallianpur:
> > > Overall this approach will unit test only service layer which is only part of the sdk code. Ideally we should test service layer, object set layer and added sdk features like pagination etc. If there is a plan to fake it , shouldn't we fake the array ?
> > > One of the approch could be to have fake_client implementation so that unit test goes through all the layers of the sdk.
> > 
> > 
> > Perhaps the purpose is not clear. I see two objectives:
> > 
> > 1. Enable unit testing of the SDK (without connecting to an array).
> > 2. Enable unit testing of a program that uses the SDK.
> > 
> > This patch targets objective #2. Specifically, the ContainerVision project needs a "fake" version of the SDK with predictable behavior and no array dependency.
> > For objective #1, I agree that faking the RESTy client layer (or equivalent) is the thing to do.
> 
> I was under impression that this is for SDK unit testing. Why would we even want to have SDK user unit testing code to be implemented here in this repo . This doesn't look right to me.

But if you look at @rgcostea response earlier, this seems like a right approach. fake package in the common-host-libs is also serving the similar purpose?

vidhutsingh:
we are talking about fake package approach and we have this implementation where actual interfaces have been modified to accommodate it . The fake package approach as quoted in the comments (ref  https://github.com/hpe-storage/common-host-libs/tree/master/storageprovider) is fine . 
I have no issue with adding completing different fake package and having all the services implementations with hard coded values in that package.  
This PR should be discarded as it doesn't make sense if are following the fake package approach.

I have provided implementation detail in this bug. 
https://nimblejira.nimblestorage.com/browse/SA-172

Also generated the fakservices with nil return value and with proper signature . Please check the bug.

PR #15, opened 2020-10-18
SA-168, the SDKs doesn't allow to pass nil param to Get request

vidhutsingh:
Problem:
User cannot specify nil param to GetVolumes request 
Fix:
Removed NIL check in service layer for Get() response. As it is not require since user may or may not pass "param" to Get() request. 
Also, there are additional generated objects due to SM repo sync as part of this PR. 

### DIFF
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -6,11 +6,12 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"strings"
+	"time"
+
 	"github.com/go-resty/resty/v2"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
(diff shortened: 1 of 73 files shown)


rkumpf (approved):

pkg/client/group_mgmt_client.go
```diff
@@ -267,7 +268,7 @@ func (client *GroupMgmtClient) listPost(
 	params *param.GetParams,
 ) (interface{}, error) {
 	// build the url
-	url := fmt.Sprintf("%s/%s/detail", client.URL, path)
+	url := fmt.Sprintf("%s/%s/details", client.URL, path)
```
rgcostea, line 271: When did this change?  And is this backwards compatible with older Nimble OS releases?

vidhutsingh: Oh no . Its wrong . Let me fix this as part of PR#16. since this is already merged . sorry about that.

PR #16, opened 2020-10-19
Added unit test for sorting, filtering,fileds and compund ops

vidhutsingh:
Fix :
Added more unit test cases around sdk features .
Fix nil pointer error . 


pkg/client/group_mgmt_client.go
```diff
@@ -232,10 +232,13 @@ func (client *GroupMgmtClient) listGetOrPost(path string, params *param.GetParam
 			fetch := "fetch"
 			wrapper := &DataWrapper{
 				Data:          params.Filter,
-				StartRow:      params.Page.StartRow,
-				EndRow:        params.Page.EndRow,
 				OperationType: &fetch,
 			}
+			if params.Page != nil && params.Page.StartRow != nil && params.Page.EndRow != nil {
```
rkumpf: Isn't it enough to just check params.Page != nil?

pkg/service/volume_service_test.go
```diff
@@ -355,6 +355,97 @@ func (suite *VolumeServiceTestSuite) TestGetVolumes() {
...
+	sf := &param.SearchFilter{
+		FieldName: nimbleos.VolumeFields.Name,
+		Operator:  param.EQUALS.String(),
+		Value:     "GetVolume",
+	}
+
+	sfilter.WithSearchFilter(sf)
+	volumes, _ := suite.volumeService.GetVolumes(sfilter)
+	if len(volumes) == 0 {
+		suite.T().Errorf("TestGetVolumes(): Unable to fetch volumes")
```
rkumpf, line 388: Can we test that the request was sent with the correct search filters?  Checking that some volumes are returned only tells us that a valid request was sent, not that it was the correct request.

These test cases are billed as unit tests, but actually they act as integration tests.  A unit test does not depend upon the behavior of another device on the network.  Could we mock the resty client to control the behavior for unit testing?

It might not be correct to fail the test when len(volumes) == 0, depending on the state of the array.  Does it only succeed if the array has a volume named "GetVolume"?

pkg/client/group_mgmt_client.go
vidhutsingh: No, since we can have nil for StartRow /EndRow.

pkg/service/volume_service_test.go
vidhutsingh: Agree. we shouldnt check on len(volumes) == 0. Actually, I should check on error instead of len(volumes) since that will validate the filter . An incorrect search filter will result in an error. 
This is unit test to validate the correctly implementation of search , sort etc .

pkg/client/group_mgmt_client.go
rkumpf: Aren't wrapper.StartRow and wrapper.EndRow initialized to nil anyway?  Why would it hurt to set them to nil?  It seems better to have fewer conditions.

pkg/service/volume_service_test.go
rkumpf: I disagree.  The SDK could, for example, omit the filter entirely from the request and this test case would still pass.

You are calling it a "unit test", but do you think that is what it is?

vidhutsingh: Disagree! 
the search filter is optional to request , here it is actually validating that it works. Since generated endpoints are not the same with and without filters to request. 
It also serves as an example for users, how to use search filters,s, or sorting with SDK.

pkg/client/group_mgmt_client.go
vidhutsingh: Its better to have this check otherwise it may lead to a crash if you try to  access nil value since it doesn't exists. 
Its fine .

rkumpf: How could a crash occur?

Consider a client program that sets params.Page.StartRow and not params.Page.EndRow.  With this implementation, we will silently drop the specified StartRow value.  By removing the extra conditions, the StartRow will be passed along to the server, which might opt to return an error if no EndRow value is given.

pkg/service/volume_service_test.go
rkumpf: How does this test validate the generated endpoint?

suneeth51 (commented):
@vidhutsingh The pagination for list output can be handled better without exposing  `startRow` and `endRow` to the user.

In Python SDK, we have completely eliminated `startRow` and `endRow` inputs and designed pagination with new attributes `limit` (pageSize) and `from_id` to handle all scenarios.
Here is the link for Python SDK reference:
https://hpe-storage.github.io/nimble-python-sdk/get_started/using/index.html#pagination 


Please take a look and let me know.

pkg/client/group_mgmt_client.go
vidhutsingh: We can go either way. I believe having this extra check or not . I didn't want to have an explicit nil value to be set for wrapper.StartRow if param.Page.StarRow is not set. Anyways i will make this change.

vidhutsingh:
> @vidhutsingh The pagination for list output can be handled better without exposing `startRow` and `endRow` to the user.
> 
> In Python SDK, we have completely eliminated `startRow` and `endRow` inputs and designed pagination with new attributes `limit` (pageSize) and `from_id` to handle all scenarios.
> Here is the link for Python SDK reference:
> https://hpe-storage.github.io/nimble-python-sdk/get_started/using/index.html#pagination
> 
> Please take a look and let me know.

Pagination is implemented with Golang SDK as well. A user has to just set the PageSize and the rest will be taken care of by SDK. Yes, we don't support "from_Id", but I feel this is not required at least right now. Maybe in the future, we can consider it.

pkg/service/volume_service_test.go
vidhutsingh: With endpoints, filter arguments are passed as key/value which gets validated on Array . if there is an error test will fail. hope this helps.

rkumpf: I think you are evaluating the test case based on your knowledge of the implementation, and I expect you are correct given the current implementation.  But I bet I could insert a bug into the implementation that causes the filters not to work at all, and this test case wouldn't catch it.

rkumpf (approved):
I maintain that these test cases are integration tests, not unit tests.  It should be possible for anyone to build this project and run the unit tests without access to a Nimble array.  That is not the case now.  I do hope this deficiency will be addressed.

vidhutsingh:
> I maintain that these test cases are integration tests, not unit tests. It should be possible for anyone to build this project and run the unit tests without access to a Nimble array. That is not the case now. I do hope this deficiency will be addressed.

You meant adding fakeprovider and unit testing. This package won't make any sense for any non-nimble customer.  I think Chaitra is working on it .

pkg/service/volume_service_test.go
rgcostea: Sorry I'm late to the conversation.  I believe @rkumpf is pointing out that we don't validate any of the possible inputs to the search filter.  For example, the operator could be something like "square_root_of" which the server doesn't support.  And the name could be something that fails to be encoded correctly.  I think it's a good practice to validate the inputs to the request, and not just validate that you get something back when the request is correctly formed.

vidhutsingh: Yes, there is no validation on search filter values we could enhance it to have validation in place. But still, the even if user supply incorrect values to search filter, the Array would throw an exception, which is returned to the caller in err form.

rgcostea: Sure. But again, that tests the array REST API server; not your code.

vidhutsingh: isn't indirectly my code getting validated :)

PR #18, opened 2020-10-22
Added SDK examples

vidhutsingh:
Descriptions :
Added SDK examples. 

### DIFF
--- a/examples/fields.go
+++ b/examples/fields.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+)
+
+func main() {
+	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
+	if err != nil {
+		fmt.Printf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
+	}
+	// set debug
+	groupService.SetDebug()
+	volSvc := groupService.GetVolumeService()
+	// Initialize volume attributes
+	var sizeField int64 = 5120
+	descriptionField := "This volume was created as part of a unit test"
+	var limitIopsField int64 = 256
+	var limitMbpsField int64 = 1
+
+	// set volume attributes
+	newVolume := &nimbleos.Volume{
+		Name:              param.NewString("TestDemo"),
+		Size:              &sizeField,
+		Description:       &descriptionField,
+		ThinlyProvisioned: param.NewBool(true),
+		Online:            param.NewBool(true),
+		LimitIops:         &limitIopsField,
+		LimitMbps:         &limitMbpsField,
+		MultiInitiator:    param.NewBool(true),
+		AgentType:         nimbleos.NsAgentTypeNone,
+	}
+
+	// create volume
+	volume, err := volSvc.CreateVolume(newVolume)
+
+	// init param
+	filter := &param.GetParams{}
+
+	// set attribute fields
+	var volAttrList = []string{
+		*nimbleos.VolumeFields.ID,
+		*nimbleos.VolumeFields.Name,
+		*nimbleos.VolumeFields.PerfpolicyName,
+	}
+	// apply attributes
+	filter.WithFields(volAttrList)
+
+	// create a filter
+	sf := &param.SearchFilter{
+		FieldName: nimbleos.VolumeFields.Name,
+		Operator:  param.EQUALS.String(),
+		Value:     "TestDemo",
+	}
+
+	filter.WithSearchFilter(sf)
+
+	volumes, err := volSvc.GetVolumes(filter)
+	fmt.Printf("%+v\n", *volumes[0])
+
+	// delete
+	volSvc.DeleteVolume(*volume.ID)
+
+	groupService.LogoutService()
+
+}
--- a/examples/fields.go
+++ b/examples/fields.go
@@ -1,3 +1,4 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
 package main
 
 import (
--- a/examples/fields.go
+++ b/examples/fields.go
@@ -13,7 +13,9 @@ func main() {
 	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
 	if err != nil {
 		fmt.Printf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
+		return
 	}
+	defer groupService.LogoutService()
 	// set debug
 	groupService.SetDebug()
 	volSvc := groupService.GetVolumeService()
@@ -38,7 +40,9 @@ func main() {
 
 	// create volume
 	volume, err := volSvc.CreateVolume(newVolume)
-
+	if err != nil {
+		fmt.Println("Failed to create volume, ", err)
+	}
 	// init param
 	filter := &param.GetParams{}
 
@@ -65,7 +69,4 @@ func main() {
 
 	// delete
 	volSvc.DeleteVolume(*volume.ID)
-
-	groupService.LogoutService()
-
 }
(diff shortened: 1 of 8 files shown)


examples/fields.go
```diff
@@ -0,0 +1,70 @@
+package main
```
rgcostea, line 2: missing copyright

vidhutsingh: thanks @rgcostea

examples/fields.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+)
+
+func main() {
+	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
+	if err != nil {
+		fmt.Printf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
```
rkumpf, line 15: Is there any point in proceeding, or should it just exit?

examples/fields.go
```diff
@@ -0,0 +1,71 @@
...
+		ThinlyProvisioned: param.NewBool(true),
+		Online:            param.NewBool(true),
+		LimitIops:         &limitIopsField,
+		LimitMbps:         &limitMbpsField,
+		MultiInitiator:    param.NewBool(true),
+		AgentType:         nimbleos.NsAgentTypeNone,
+	}
+
+	// create volume
+	volume, err := volSvc.CreateVolume(newVolume)
```
rkumpf, line 42: Include error checks?

examples/fields.go
```diff
@@ -0,0 +1,71 @@
...
+	// Initialize volume attributes
+	var sizeField int64 = 5120
+	descriptionField := "This volume was created as part of a unit test"
+	var limitIopsField int64 = 256
+	var limitMbpsField int64 = 1
+
+	// set volume attributes
+	newVolume := &nimbleos.Volume{
+		Name:              param.NewString("TestDemo"),
+		Size:              &sizeField,
```
rkumpf, line 31: Why not use param.NewInt64()?

examples/filters.go
```diff
@@ -0,0 +1,42 @@
...
+	// set debug
+	groupService.SetDebug()
+
+	// get volume service instance
+	volSvc := groupService.GetVolumeService()
+
+	// init param
+	sfilter := &param.GetParams{}
+	// set search filter
+	f := &param.SearchFilter{
```
rkumpf, line 29: It might be helpful to show how to specify multiple filters.

examples/initiator_group.go
```diff
@@ -0,0 +1,104 @@
...
+
+		// create acl
+		_, err = aclSvc.CreateAccessControlRecord(acl)
+		if err != nil {
+			fmt.Println(err)
+		}
+	}
+
+	// cleaup
+	volume, _ = volSvc.GetVolumeById(*volume.ID)
```
rkumpf: Is this statement relevant?

examples/initiator_group.go
```diff
@@ -0,0 +1,104 @@
...
+		AccessProtocol:  nimbleos.NsAccessProtocolIscsi,
+		IscsiInitiators: initiatorList,
+	}
+
+	// create initiator group
+	igroup, _ = igroupSvc.CreateInitiatorGroup(igroup)
+
+	// create a new volume and assigne acl
+	volume, _ := volSvc.CreateVolume(newVolume)
+	if volume != nil {
```
rkumpf, line 76: Is it better to check for volume != nil or for err == nil?

examples/restore.go
```diff
@@ -0,0 +1,89 @@
...
+		LimitIops:         &limitIopsField,
+		LimitMbps:         &limitMbpsField,
+		MultiInitiator:    param.NewBool(true),
+		AgentType:         nimbleos.NsAgentTypeNone,
+	}
+
+	// create volume
+	volume, err := volSvc.CreateVolume(newVolume)
+	if volume != nil {
+		volcollname := &nimbleos.VolumeCollection{
```
rkumpf: Change volcollname to volcoll?

examples/restore.go
```diff
@@ -0,0 +1,89 @@
...
+			fmt.Printf("Failed to remove %s volume from volume collection", *volume.ID)
+			return
+		}
+		// delete volcoll
+		groupService.GetVolumeCollectionService().DeleteVolumeCollection(*volcoll.ID)
+		volSvc.DeleteVolume(*volume.ID)
+
+	}
+	// logout
+	groupService.LogoutService()
```
rkumpf: Should we use a "defer" statement right after initialization so the logout occurs even when we hit the various "return" statements?

examples/restore.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// create volume
+	volume, err := volSvc.CreateVolume(newVolume)
+	if volume != nil {
+		volcollname := &nimbleos.VolumeCollection{
+			Name: param.NewString("TestVolcoll"),
+		}
+		volcoll, err := groupService.GetVolumeCollectionService().CreateVolumeCollection(volcollname)
+		if err != nil {
+			fmt.Println("Failed to get volume collection")
```
rkumpf: "get" -> "create"

pkg/service/volume_service.go
```diff
@@ -71,6 +72,19 @@ func (svc *VolumeService) GetVolumeById(id string) (*nimbleos.Volume, error) {
 	return volumeResp, nil
 }
 
+// GetVolumeStats - method returns a pointer to "Stats"
+func (svc *VolumeService) GetVolumeStats(id string) (*nimbleos.Stat, error) {
```
rkumpf: How might one provide filters to this call?  For example, "GET /v1/volumes/<id>/stats?sensors=reads,writes".

examples/initiator_group.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	// cleaup
+	volume, _ = volSvc.GetVolumeById(*volume.ID)
+	updateVol := &nimbleos.Volume{
+		Online: param.NewBool(false),
+	}
+	// set volume offline
+	volume, _ = volSvc.UpdateVolume(*volume.ID, updateVol)
```
rkumpf, line 94: Earlier we had talked about providing an example of how a client program could handle a 202 Accepted response when WaitForJobs==false.  Is that planned?

examples/fields.go
vidhutsingh: It shd exit

examples/fields.go
vidhutsingh: sure

examples/fields.go
vidhutsingh: We can use it . lets keep it for now .

examples/filters.go
vidhutsingh: we can add a different example, not in this file

examples/initiator_group.go
vidhutsingh: Not required

examples/restore.go
vidhutsingh: No, it should be volcollName

examples/restore.go
vidhutsingh: Make sense .

examples/restore.go
vidhutsingh: ok

pkg/service/volume_service.go
vidhutsingh: Filter is not supported for now .

examples/initiator_group.go
vidhutsingh: You just need to parse the error code and filter out job id (by using regex ) and invoke GetJobId() . I will create a separate example for it . Not here

examples/fields.go
rkumpf: The same comment applies to several of the examples.

examples/fields.go
rkumpf: The same comment applies to several of the examples.

rkumpf (dismissed):
It is important that best practices are exemplified, because people will definitely copy and paste from these examples directly into their client programs.

rkumpf (approved):

PR #20, opened 2020-10-29
Provided fakservice support for unit testing

vidhutsingh:
Bug : SA-172
Implementation:
Added the framework to support fake services for unit testing. Initially added volume_service and pool_service. 
User has to call fakeservice package instead of service package for NewNsGroupService() 

Example 
```
        package main

import (
        "fmt"

        "github.hpe.com/nimble-dcs/golang-sdk/pkg/client/v1/nimbleos"
        "github.hpe.com/nimble-dcs/golang-sdk/pkg/param"
        "github.hpe.com/nimble-dcs/golang-sdk/pkg/fakeservice"
)

func main() {
        groupService, err := fakeservice.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
        if err != nil {
                fmt.Printf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
                return
        }
        // set debug
        //groupService.SetDebug()
        defer groupService.LogoutService()
        // get  volume service instance
        volSvc := groupService.GetVolumeService()
        // Initialize volume attributes
        var sizeField int64 = 5120
        var limitMbpsField int64 = 1

        // set volume attributes
        newVolume := &nimbleos.Volume{
                ID:                param.NewString("0677654354e496314900000000000000000000000b"),
                Name:              param.NewString("TestDemo1"),
                Size:              &sizeField,
                Description:       &descriptionField,
                ThinlyProvisioned: param.NewBool(true),
                Online:            param.NewBool(true),
                LimitIops:         &limitIopsField,
                LimitMbps:         &limitMbpsField,
                MultiInitiator:    param.NewBool(true),
                AgentType:         nimbleos.NsAgentTypeNone,
        }

        // create volume
        volume, err := volSvc.CreateVolume(newVolume)

        if err != nil {
                fmt.Printf("Failed to create volume, err ", err)
                return
        }
        fmt.Printf("vol : %+v", volume)
        // delete volume, cleanup
        volSvc.DeleteVolume(*volume.ID)

        fmt.Printf("error : %v", err)
        _, err = volSvc.GetVolumeById(*volume.ID)
        fmt.Printf("error : %v", err)

        poolSvc := groupService.GetPoolService()
        p, _ := poolSvc.GetPools(nil)
        fmt.Printf("pool %v", p)

}
```

### DIFF
--- a/pkg/fakeservice/group_service_factory.go
+++ b/pkg/fakeservice/group_service_factory.go
@@ -1,0 +1,38 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+
+package fakeservice
+
+// NsGroupService type
+type NsGroupService struct {
+	ip string
+	// Declare all the supported services
+	volumeService *VolumeService
+	poolService   *PoolService
+}
+
+// NewNsGroupService - initializes NsGroupService
+func NewNsGroupService(ip, username, password, apiVersion string, synchronous bool) (gs *NsGroupService, err error) {
+
+	return &NsGroupService{ip: ip}, nil
+}
+
+// LogoutService - delete the session token
+func (gs *NsGroupService) LogoutService() error {
+	return nil
+}
+
+// GetVolumeService - returns service of a type *VolumeService
+func (gs *NsGroupService) GetVolumeService() (vs *VolumeService) {
+	if gs.volumeService == nil {
+		gs.volumeService = NewVolumeService(gs)
+	}
+	return gs.volumeService
+}
+
+// GetPoolService - returns service of a type *PoolService
+func (gs *NsGroupService) GetPoolService() (vs *PoolService) {
+	if gs.poolService == nil {
+		gs.poolService = NewPoolService(gs)
+	}
+	return gs.poolService
+}
--- a/pkg/fakeservice/pool_service.go
+++ b/pkg/fakeservice/pool_service.go
@@ -5,6 +5,8 @@ package fakeservice
 // Pool Service - Manage pools. Pools are an aggregation of arrays.
 
 import (
+	"fmt"
+
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 )
@@ -12,6 +14,7 @@ import (
 // PoolService type
 type PoolService struct {
 	// create a map
+	PoolMemory map[string]*nimbleos.Pool
 }
 
 // NewPoolService - method to initialize "PoolService"
@@ -22,32 +25,59 @@ func NewPoolService(gs *NsGroupService) *PoolService {
 
 // GetPools - method returns a array of pointers of type "Pools"
 func (svc *PoolService) GetPools(params *param.GetParams) ([]*nimbleos.Pool, error) {
-	return nil, nil
+	var pools []*nimbleos.Pool
+	for _, v := range svc.PoolMemory {
+		pools = append(pools, v)
+	}
+	return pools, nil
 }
 
 // CreatePool - method creates a "Pool"
 func (svc *PoolService) CreatePool(obj *nimbleos.Pool) (*nimbleos.Pool, error) {
-	return nil, nil
+	if svc.PoolMemory == nil {
+		svc.PoolMemory = make(map[string]*nimbleos.Pool)
+	}
+	if _, ok := svc.PoolMemory[*obj.ID]; !ok {
+		svc.PoolMemory[*obj.ID] = obj
+	} else {
+		return nil, fmt.Errorf("object already exists, duplicate pool %v", obj.Name)
+	}
+	return obj, nil
 }
 
 // UpdatePool - method modifies  the "Pool"
 func (svc *PoolService) UpdatePool(id string, obj *nimbleos.Pool) (*nimbleos.Pool, error) {
-	return nil, nil
+	if _, ok := svc.PoolMemory[id]; ok {
+		svc.PoolMemory[id] = obj
+	} else {
+		return nil, fmt.Errorf("object doesn't exist, failed to update pool %v", obj.Name)
+	}
+	return obj, nil
 }
 
 // GetPoolById - method returns a pointer to "Pool"
 func (svc *PoolService) GetPoolById(id string) (*nimbleos.Pool, error) {
+	if _, ok := svc.PoolMemory[id]; ok {
+		return svc.PoolMemory[id], nil
+	}
 	return nil, nil
 }
 
 // GetPoolByName - method returns a pointer "Pool"
 func (svc *PoolService) GetPoolByName(name string) (*nimbleos.Pool, error) {
+	for _, v := range svc.PoolMemory {
+		if *v.Name == name {
+			return v, nil
+		}
+	}
 	return nil, nil
 }
 
 // DeletePool - deletes the "Pool"
 func (svc *PoolService) DeletePool(id string) error {
-
+	if _, ok := svc.PoolMemory[id]; ok {
+		delete(svc.PoolMemory, id)
+	}
 	return nil
 }
 
@@ -60,6 +90,8 @@ func (svc *PoolService) DeletePool(id string) error {
 //       force - Forcibly merge the specified pool into target pool.
 
 func (svc *PoolService) MergePool(id string, targetPoolId string, force *bool) (*nimbleos.NsPoolMergeReturn, error) {
-
+	if _, ok := svc.PoolMemory[id]; ok {
+		delete(svc.PoolMemory, id)
+	}
 	return nil, nil
 }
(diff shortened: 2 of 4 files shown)


ckallianpur:
Looked a bit more at the way the kubernetes fake package is implemented. It provides a way to set a object before the test and verify if all the code paths are covered. For example like this
`	secretObj := core_v1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Namespace: "namespace",
			Name:      "secret",
		},
	}
	client := KubeClient{}
	client.Client = fake.NewSimpleClientset(&secretObj)`

NewSimpleClientset here returns a clientset that will respond with the provided objects. It's backed by a very simple object tracker that processes creates, updates and deletions as-is, without applying any validations and/or defaults. 

Could we update the logic in the fake package here to allow something similar. If at all we just hardcode the return values, say from GetVolume or GetPools, it will be tailored for a client using it and might not satisfy the requirements of other clients using it.

vidhutsingh:
> Looked a bit more at the way the kubernetes fake package is implemented. It provides a way to set a object before the test and verify if all the code paths are covered. For example like this
> ` secretObj := core_v1.Secret{ ObjectMeta: metav1.ObjectMeta{ Namespace: "namespace", Name: "secret", }, } client := KubeClient{} client.Client = fake.NewSimpleClientset(&secretObj)`
> 
> NewSimpleClientset here returns a clientset that will respond with the provided objects. It's backed by a very simple object tracker that processes creates, updates and deletions as-is, without applying any validations and/or defaults.
> 
> Could we update the logic in the fake package here to allow something similar. If at all we just hardcode the return values, say from GetVolume or GetPools, it will be tailored for a client using it and might not satisfy the requirements of other clients using it.

With the current implementation, you can do the same thing.  
func (svc *VolumeService) CreateVolume(obj *nimbleos.Volume) (*nimbleos.Volume, error)  - this is signature of the CreateVolume(). As you can see it accepts *nimbleos.Volume as an input argument, by using it you can create a map of volume_id and nimbleos.Volume , store it in VolumeService struct. Now you can track that volume via remaining functions like GetVolume etc .

ckallianpur:
> > Looked a bit more at the way the kubernetes fake package is implemented. It provides a way to set a object before the test and verify if all the code paths are covered. For example like this
> > ` secretObj := core_v1.Secret{ ObjectMeta: metav1.ObjectMeta{ Namespace: "namespace", Name: "secret", }, } client := KubeClient{} client.Client = fake.NewSimpleClientset(&secretObj)`
> > NewSimpleClientset here returns a clientset that will respond with the provided objects. It's backed by a very simple object tracker that processes creates, updates and deletions as-is, without applying any validations and/or defaults.
> > Could we update the logic in the fake package here to allow something similar. If at all we just hardcode the return values, say from GetVolume or GetPools, it will be tailored for a client using it and might not satisfy the requirements of other clients using it.
> 
> With the current implementation, you can do the same thing.
> func (svc *VolumeService) CreateVolume(obj *nimbleos.Volume) (*nimbleos.Volume, error) - this is signature of the CreateVolume(). As you can see it accepts *nimbleos.Volume as an input argument, by using it you can create a map of volume_id and nimbleos.Volume , store it in VolumeService struct. Now you can track that volume via remaining functions like GetVolume etc .

But that is not in the fake package

vidhutsingh:
> > > Looked a bit more at the way the kubernetes fake package is implemented. It provides a way to set a object before the test and verify if all the code paths are covered. For example like this
> > > ` secretObj := core_v1.Secret{ ObjectMeta: metav1.ObjectMeta{ Namespace: "namespace", Name: "secret", }, } client := KubeClient{} client.Client = fake.NewSimpleClientset(&secretObj)`
> > > NewSimpleClientset here returns a clientset that will respond with the provided objects. It's backed by a very simple object tracker that processes creates, updates and deletions as-is, without applying any validations and/or defaults.
> > > Could we update the logic in the fake package here to allow something similar. If at all we just hardcode the return values, say from GetVolume or GetPools, it will be tailored for a client using it and might not satisfy the requirements of other clients using it.
> > 
> > 
> > With the current implementation, you can do the same thing.
> > func (svc *VolumeService) CreateVolume(obj *nimbleos.Volume) (*nimbleos.Volume, error) - this is signature of the CreateVolume(). As you can see it accepts *nimbleos.Volume as an input argument, by using it you can create a map of volume_id and nimbleos.Volume , store it in VolumeService struct. Now you can track that volume via remaining functions like GetVolume etc .
> 
> But that is not in the fake package
I am quoting from fakeservice only. It just that you have to add map in VolumeService struct .

ckallianpur:
@vidhutsingh - The implementation looks good but I am still not following how without creating an interface this will work. Suppose I have a method which is taking *service.NsGroupService as an argument, it wont accept it if I pass in  *fakeservice.NsGroupService? Instead if both fake and actual ones have an interface then, if the clients use the interface the test code can pass in fake and real code can pass in actual implementation. Similar to the way the common-host-libs implements it.

ckallianpur:
@vidhutsingh  I am happy to make those changes.

vidhutsingh:
> @vidhutsingh - The implementation looks good but I am still not following how without creating an interface this will work. Suppose I have a method which is taking *service.NsGroupService as an argument, it wont accept it if I pass in *fakeservice.NsGroupService? Instead if both fake and actual ones have an interface then, if the clients use the interface the test code can pass in fake and real code can pass in actual implementation. Similar to the way the common-host-libs implements it.

Why is that difficult to conditionally invoke in your code 
groupService, err := fakeservice.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
vs 
groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)


You dont have to change anything else in your code. 
The current implementation suffices the requirement for thirdparty app to do  unit testing on Pool and Volume services . I have given example as well . please show me your code . thanks

rkumpf:
The issue is that both the real and fake services need to have the same type, so they can be used interchangeably as arguments to a function that uses the service.  For example, this code needs to compile:

    var groupService *service.NsGroupService
    var err error

    groupService, err = service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
    groupService, err = fakeservice.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)

This requirement drives a need to define an interface for both versions to implement.  (And once we have an interface, there isn't much reason to use a pointer anymore.)

vidhutsingh:
> The issue is that both the real and fake services need to have the same type, so they can be used interchangeably as arguments to a function that uses the service. For example, this code needs to compile:
> 
> ```
> var groupService *service.NsGroupService
> var err error
> 
> groupService, err = service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
> groupService, err = fakeservice.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
> ```
> 
> This requirement drives a need to define an interface for both versions to implement. (And once we have an interface, there isn't much reason to use a pointer anymore.)

You can also define your own interface as per your need and pass it around , we don't need to make such changes in the sdk . 
It is more part of your application requirement than SDK and we shouldn't have third party application logic here .
Can i see your application code ?

rkumpf:
You can find an example of code using the Go client SDK here: https://github.hpe.com/nimble-dcs/array-exporter/blob/master/pkg/collector/volume.go.

I think you are suggesting the SDK doesn't need to provide interfaces or fake implementations because the calling code could provide them instead.  However, it means the work needs to be repeated in all the calling applications.  Your viewpoint seems unconventional in the Go realm.

vidhutsingh:
r Go client SDK here: https://github.hpe.com/nimble-dcs/array-exporter/blob/master/pkg/collector/volume.go.
> 
> I think you are suggesting the SDK doesn't need to provide interfaces or fake implementations because the calling code could provide them instead. However, it means the work needs to be repeated in all the calling applications. Your viewpoint seems unconventional in the Go realm.

Fake implementation is already provided you just have to write your own wrapper package around fakservice and service with application code.

ckallianpur:
> r Go client SDK here: https://github.hpe.com/nimble-dcs/array-exporter/blob/master/pkg/collector/volume.go.
> 
> > I think you are suggesting the SDK doesn't need to provide interfaces or fake implementations because the calling code could provide them instead. However, it means the work needs to be repeated in all the calling applications. Your viewpoint seems unconventional in the Go realm.
> 
> Fake implementation is already provided you just have to write your own wrapper package around fakservice and service with application code.

@vidhutsingh - Based on the discussions in email with George, the interface needs to be added in SDK itself. Thoughts?

ckallianpur:
> > r Go client SDK here: https://github.hpe.com/nimble-dcs/array-exporter/blob/master/pkg/collector/volume.go.
> > > I think you are suggesting the SDK doesn't need to provide interfaces or fake implementations because the calling code could provide them instead. However, it means the work needs to be repeated in all the calling applications. Your viewpoint seems unconventional in the Go realm.
> > 
> > 
> > Fake implementation is already provided you just have to write your own wrapper package around fakservice and service with application code.
> 
> @vidhutsingh - Based on the discussions in email with George, the interface needs to be added in SDK itself. Thoughts?

Sample implementation with interface https://github.com/hpe-storage/nimble-golang-sdk/pull/14/files

PR #21, opened 2020-11-10
Adding workflow tests for Volume, IG, ACR

prachiruparelia-hpe:


workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+)
+
+var defaultVolumeName = "DefaultVolumeTest"
+var defaultInitiatorGrp = "DefaultInitiatorgrpTest"
+var defaultVolColName = "DefaultVolColTest"
+
+func config() *service.NsGroupService {
+	groupService, err := service.NewNsGroupService("10.21.1.216", "admin", "admin")
```
rkumpf: Sensitive data must not be submitted to any code repository.

workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		fmt.Errorf("NewGroupService(): unable to connect to group, err: %v", err.Error())
+	}
+
+	// set debug
+	groupService.SetDebug()
+	// suite.groupService = groupService
```
rkumpf: Dead code?

workflows/access_control_records_test.go
```diff
@@ -0,0 +1,73 @@
...
+	createDefaultInitiatorGrp(groupService)
+}
+
+func (suite *ACRWorkflowSuite) TearDownSuite() {
+	err := suite.acrService.DeleteAccessControlRecord(*suite.acr.ID)
+	if err != nil {
+		suite.T().Errorf("Access Control Record deletion failed: %v", err.Error())
+	}
+	deleteVolume(suite.groupService, defaultVolumeName)
+	deleteInitiatorGroup(suite.groupService, defaultInitiatorGrp)
```
rkumpf: Add suite.groupService.LogoutService()?

workflows/access_control_records_test.go
```diff
@@ -0,0 +1,73 @@
...
+	createDefaultVolume(groupService)
+	createDefaultInitiatorGrp(groupService)
+}
+
+func (suite *ACRWorkflowSuite) TearDownSuite() {
+	err := suite.acrService.DeleteAccessControlRecord(*suite.acr.ID)
+	if err != nil {
+		suite.T().Errorf("Access Control Record deletion failed: %v", err.Error())
+	}
+	deleteVolume(suite.groupService, defaultVolumeName)
```
rkumpf: This isn't symmetric with the createDefaultVolume(groupService) statement.  Either this module should provide the name for both create and delete, or the common module should manage it with a deleteDefaultVolume() function.

workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+var defaultVolumeName = "DefaultVolumeTest"
+var defaultInitiatorGrp = "DefaultInitiatorgrpTest"
+var defaultVolColName = "DefaultVolColTest"
+
+func config() *service.NsGroupService {
+	groupService, err := service.NewNsGroupService("10.21.1.216", "admin", "admin")
+	if err != nil {
+		fmt.Errorf("NewGroupService(): unable to connect to group, err: %v", err.Error())
```
rkumpf: Errors in this file don't seem to cause test suite failures.

workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"fmt"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+)
+
+var defaultVolumeName = "DefaultVolumeTest"
+var defaultInitiatorGrp = "DefaultInitiatorgrpTest"
+var defaultVolColName = "DefaultVolColTest"
```
rkumpf: I think the preferred abbreviation is VolColl.

workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	// set debug
+	groupService.SetDebug()
+	// suite.groupService = groupService
+	// suite.volumeService = groupService.GetVolumeService()
+	return groupService
+}
+
+func createDefaultVolume(groupService *service.NsGroupService) (*nimbleos.Volume, error) {
```
rkumpf: Should the parameter be a volumeService?

workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+	volcolService := groupService.GetVolumeCollectionService()
+	_, err := volcolService.CreateVolumeCollection(newVolCol)
+	if err != nil {
+		fmt.Errorf("Volume Collection creation failed: %v", err.Error())
+	}
+}
+
+func deleteVolume(groupService *service.NsGroupService, volName string) {
+	volSvc := groupService.GetVolumeService()
+	volObj, _ := volSvc.GetVolumeByName(volName)
```
rkumpf: Some errors (other than Not Found) are worth reporting.

workflows/access_control_records_test.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+package workflows
```
rkumpf: Should the package name be singular?

workflows/access_control_records_test.go
```diff
@@ -0,0 +1,73 @@
...
+	suite.groupService = groupService
+	suite.volumeService = groupService.GetVolumeService()
+	suite.igService = groupService.GetInitiatorGroupService()
+	suite.acrService = groupService.GetAccessControlRecordService()
+	createDefaultVolume(groupService)
+	createDefaultInitiatorGrp(groupService)
+}
+
+func (suite *ACRWorkflowSuite) TearDownSuite() {
+	err := suite.acrService.DeleteAccessControlRecord(*suite.acr.ID)
```
rkumpf: Might acr be nil here, causing a panic?

workflows/common.go
```diff
@@ -0,0 +1,92 @@
...
+
+func createDefaultInitiatorGrp(igSvc *service.InitiatorGroupService) {
+	newIG := &nimbleos.InitiatorGroup{
+		Name:           param.NewString(defaultInitiatorGrpName),
+		Description:    param.NewString("Workflow initiator group"),
+		AccessProtocol: nimbleos.NsAccessProtocolIscsi,
+	}
+	_, err := igSvc.CreateInitiatorGroup(newIG)
+	if err != nil {
+		fmt.Errorf("Initiator group creation failed with error: %v", err.Error())
```
rkumpf: I suggest returning the errors from these functions so the callers can verify success.
These fmt.Errorf statements create error structures which are not stored anywhere, so they don't seem to have any effect.

workflows/access_control_records_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (suite *ACRWorkflowSuite) TestCreateACR() {
+	vol, _ := suite.volumeService.GetVolumeByName(defaultVolumeName)
+	ig, _ := suite.igService.GetInitiatorGroupByName(defaultInitiatorGrpName)
+	newACR := &nimbleos.AccessControlRecord{
+		InitiatorGroupId: ig.ID,
+		VolId:            vol.ID,
+	}
+	acr, err := suite.acrService.CreateAccessControlRecord(newACR)
+	if err != nil {
```
rkumpf: Was use of the assert framework considered?  It might help streamline the error checks.

workflows/volume_test.go
```diff
@@ -0,0 +1,200 @@
...
+	newVolume := &nimbleos.Volume{
+		Size:      &sizeIncField,
+		LimitIops: &limitIopsField,
+		LimitMbps: &limitMbpsField,
+	}
+	vol, _ := suite.volumeService.GetVolumeByName(volumeName)
+	if vol != nil {
+		_, err := suite.volumeService.UpdateVolume(*vol.ID, newVolume)
+		if err != nil {
+			suite.T().Errorf("Volume modify didnt go well %v:", err.Error())
```
rkumpf: Move the colon before %v.

workflows/access_control_records_test.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+package workflow
```
rkumpf: The directory name should match the package name.

rkumpf (commented):
I'm curious about the purpose of these tests.  Are they meant to test the array itself?

workflows/access_control_records_test.go
prachiruparelia-hpe: @rkumpf I am seeing strange issue when I try to use assert. Example, when I replace below code:
If err ! = nil {
   suite.T().Errorf("Volume creation failed with error: %v", err.Error())
}
with:
assert.Nil(suite.T(), err, "Volume creation failed with error: %v", err.Error())
If err value is not nil its failing as expected. 
But if err is nil, that is if volume got created successfully, it gives nil pointer exception on the assert line.
Not sure if its me, or there is a bug with assert package from go.

rkumpf: Yes, that makes sense.  The parameters are evaluated before the function is called, so err.Error() would deference a nil pointer in that case.  You could replace err.Error() with just err.

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+	suite.deleteInitiatorGroup(initiatorGroupName)
+	err := deleteDefaultVolume(suite.volumeService)
+	assert.Nilf(suite.T(), err, "Unable to delete default volume, err: %v", err)
+}
+
+func (suite *IGWorkflowSuite) deleteInitiatorGroup(igName string) {
+	ig, _ := suite.initiatorGrpService.GetInitiatorGroupByName(igName)
+	if ig != nil {
+		err := suite.initiatorGrpService.DeleteInitiatorGroup(*ig.ID)
+		assert.Nilf(suite.T(), err, "Unable to delete Initiator group, err: %v", igName)
```
rkumpf: Substitution parameter mismatch

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+	var initiatorList []*nimbleos.NsISCSIInitiator
+	initiatorList = append(initiatorList, iscsiInitiator)
+	newIG := &nimbleos.InitiatorGroup{
+		Name:            param.NewString("TestIGIscsiInvalid"),
+		Description:     param.NewString("Workflow initiator group"),
+		AccessProtocol:  nimbleos.NsAccessProtocolIscsi,
+		IscsiInitiators: initiatorList,
+	}
+	_, err := suite.initiatorGrpService.CreateInitiatorGroup(newIG)
+	assert.NotNil(suite.T(), err, "Initiator group cretaion with invalid iscsi initiators should have failed")
```
rkumpf: Typo *creation

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+	var desc *string = param.NewString("Workflow tests for initiator group")
+	updateIG := &nimbleos.InitiatorGroup{
+		Description: desc,
+	}
+	ig, _ := suite.initiatorGrpService.GetInitiatorGroupByName(initiatorGroupName)
+	if ig != nil {
+		suite.initiatorGrpService.UpdateInitiatorGroup(*ig.ID, updateIG)
+	}
+	ig, _ = suite.initiatorGrpService.GetInitiatorGroupByName(initiatorGroupName)
+	assert.Equalf(suite.T(), *desc, *ig.Description, "Unable to update Initiatot Group, err: %v", ig.Description)
```
rkumpf: Remove "err".

Typo *Initiator

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	var initiatorList []*nimbleos.NsISCSIInitiator
+	initiatorList = append(initiatorList, iscsiInitiator)
+	updateIG := &nimbleos.InitiatorGroup{
+		IscsiInitiators: initiatorList,
+	}
+	currentIG, _ := suite.initiatorGrpService.GetInitiatorGroupByName("TestIGIscsi")
+	updatedIG, err := suite.initiatorGrpService.UpdateInitiatorGroup(*currentIG.ID, updateIG)
+	assert.Nilf(suite.T(), err, "Modifying IP address of initiator failed: %v", err)
+	assert.Equalf(suite.T(), *ipAdd, *updateIG.IscsiInitiators[0].IpAddress, "Updating iscsi initiatos failed: %v", updatedIG.IscsiInitiators[0].IpAddress)
```
rkumpf: Typo *initiators

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+	var desc *string = param.NewString("Workflow initiator group")
+	newIG := &nimbleos.InitiatorGroup{
+		Name:           param.NewString(initiatorGroupName),
+		Description:    desc,
+		AccessProtocol: nimbleos.NsAccessProtocolIscsi,
+	}
+	_, err := suite.initiatorGrpService.CreateInitiatorGroup(newIG)
+	assert.Nilf(suite.T(), err, "Unable to create initiator group, err: %v", err)
+	ig, _ := suite.initiatorGrpService.GetInitiatorGroupByName(initiatorGroupName)
+	assert.Equalf(suite.T(), *desc, *ig.Description, "Initiator group creation does not have expected description: %v", desc)
```
rkumpf: We shouldn't need to include the IP address in the message.  The test framework will show us the expected and actual values.  This comment applies to all the assert.Equalf statements in this patch.  They can be changed to assert.Equal.

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+	iscsiInitiator := &nimbleos.NsISCSIInitiator{
+		IpAddress: ipAdd,
+		Label:     param.NewString("NewLabel"),
+	}
+	var initiatorList []*nimbleos.NsISCSIInitiator
+	initiatorList = append(initiatorList, iscsiInitiator)
+	updateIG := &nimbleos.InitiatorGroup{
+		IscsiInitiators: initiatorList,
+	}
+	currentIG, _ := suite.initiatorGrpService.GetInitiatorGroupByName("TestIGIscsi")
```
rkumpf: I wonder if each test case should be self-contained, rather than relying on the result from another test case.  For example, this test case could perform a Create, Modify, and Delete.

workflow/common.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+)
+
+var defaultVolumeName = "DefaultVolumeTest"
+var defaultInitiatorGrpName = "DefaultInitiatorgrpTest"
+var defaultVolCollName = "DefaultVolCollTest"
+
+func config() (*service.NsGroupService, error) {
+	groupService, err := service.NewNsGroupService("", "admin", "admin", "v1", true)
```
rkumpf: What does it mean for the address to be empty?  Is customization required?  An alternative is to read the connection details from the environment.

Passwords should never be submitted to source repositories.

workflow/common.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return groupService, err
+	}
+	defer groupService.LogoutService()
```
rkumpf: This causes a logout to be performed at the end of this function.  I don't think that is wanted.

workflow/initiator_group_test.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+var initiatorGroupName = "InitiatorGroupTest"
```
rkumpf: Should it be const instead of var?  This comment also applies in a few other places.

rkumpf (commented):
Many of my comments are minor and could be addressed in a subsequent patch.  A couple, like the password one, are needed before merging.

Overall this looks good to me, although I am not familiar with the purpose of these tests.

workflow/initiator_group_test.go
prachiruparelia-hpe: Makes sense. Will take care of it in my next PR.

workflow/common.go
```diff
@@ -0,0 +1,72 @@
+package workflow
```
rkumpf: Several files are missing copyright statements.

rkumpf (approved):

workflow/access_control_records_test.go
```diff
@@ -0,0 +1,71 @@
...
+func (suite *ACRWorkflowSuite) TearDownSuite() {
+	if suite.acr != nil {
+		err := suite.acrService.DeleteAccessControlRecord(*suite.acr.ID)
+		assert.Nilf(suite.T(), err, "Unable to delete Access control record, err: %v", err)
+	}
+	err := deleteDefaultVolume(suite.volumeService)
+	assert.Nilf(suite.T(), err, "Unable to delete default volume, err: %v", err)
+	err = deleteDefaultInitiatorGroup(suite.igService)
+	assert.Nilf(suite.T(), err, "Unable to delete default initiator group, err: %v", err)
+}
```
rkumpf: I suggest adding a groupService Logout statement here.

rkumpf (approved):
See the failed check, and follow the instructions to properly sign off all the commits.

https://github.com/hpe-storage/nimble-golang-sdk/pull/21/checks?check_run_id=1418852220

workflow/access_control_records_test.go
prachiruparelia-hpe: Updated.

vidhutsingh (changes requested):
workflow name is looking bit off , shall we change it to sdktesting ?

vidhutsingh (approved):

PR #23, opened 2020-11-19
Added workflow tests for Group and Pool

kousalya-hpe:
Added workflow tests for Group and Pool

### DIFF
--- a/pkg/workflow/group_pool_wf_test.go
+++ b/pkg/workflow/group_pool_wf_test.go
@@ -1,0 +1,335 @@
+// Group tests
+// ------------
+//    1. Get group info
+//    2. validate group merge
+//    3. Do group merge
+//    4. modify default_iscsi_target_scope
+// Pool tests
+// -----------
+//   1. Get pool details - get pools name
+//   2. Merge two pools
+//   3. Edit Pool - remove array from pool
+//   4. Create a new pool with available / above removed pool
+//   5. Change pool name
+//   6. Create volume in new pool
+//   7. Delete pool when volumes are present
+//       Should fail, validate error msg
+//   8. Delete volume from new pool
+//   9. Delete pool
+//       Should be deleted
+
+package workflow
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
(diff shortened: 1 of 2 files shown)


pkg/workflow/group_pool_wf_test.go
```diff
@@ -0,0 +1,335 @@
...
+	"time"
+
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+var sourceArrayIP = "1.1.1.1"
```
prachiruparelia-hpe: All these would be better as const.

pkg/workflow/group_pool_wf_test.go
```diff
@@ -0,0 +1,335 @@
...
+		ArrayList: updateArrayList,
+	}
+	_, err := suite.poolService.UpdatePool(defaultPoolID, updatePoolInput)
+	assert.Nilf(suite.T(), err, "Pool Array list update failed")
+	fmt.Printf("Updated pool array list\n")
+	time.Sleep(5 * time.Second)
+}
+
+// Check if the element is present in the provided list/array/slice
+func Contains(slice []string, val string) (int, bool) {
```
prachiruparelia-hpe: I looked at volume_service_test as an example, and the helper functions are starting with lower case. So contains would probably be consistent with other test files.

pkg/workflow/group_pool_wf_test.go
```diff
@@ -0,0 +1,335 @@
...
+	fmt.Printf("Error %+v\n", err.Error())
+	assert.NotNil(suite.T(), err, "Pool Deletion expected to be failed")
+
+	// Delete Volume
+	suite.deleteVolume(volumeName)
+	time.Sleep(5 * time.Second)
+
+	// Delete Pool
+	err = suite.poolService.DeletePool(newPoolID)
+	fmt.Printf("Error %+v\n", err)
```
prachiruparelia-hpe: You probably dont need err print here since Nilf is going to print it for you.

pkg/workflow/group_pool_wf_test.go
```diff
@@ -0,0 +1,335 @@
+// Group tests
```
prachiruparelia-hpe: Package name and folder name needs updating to test.

PR #24, opened 2020-11-20
Sa 176 - Added DeleteVolume function and renamed the existing DeleteVolume to DeestroyVolumee 

vidhutsingh:
Fix : 
Added DeleteVolume which doesnt make the volume offline and tries to delete the volume . It is expected from user to ensure volume is offline . wherein DestroyVolume makes the volume offline prior to delete .


pkg/service/volume_service.go
```diff
@@ -145,10 +145,10 @@ func (svc *VolumeService) OfflineVolume(id string, force bool) (*nimbleos.Volume
 
 }
 
-// DeleteVolume - deletes the volume
-func (svc *VolumeService) DeleteVolume(id string) error {
+// DestroyVolume - deletes the volume
```
rkumpf: The description does not distinguish it from DeleteVolume.

pkg/service/snapshot_service.go
```diff
@@ -72,13 +72,13 @@ func (svc *SnapshotService) GetSnapshotById(id string) (*nimbleos.Snapshot, erro
 	return snapshotResp, nil
 }
 
-// GetSnapshotByName - method returns a pointer "Snapshot"
-func (svc *SnapshotService) GetSnapshotByName(name string) (*nimbleos.Snapshot, error) {
+// GetSnapshotByName - method returns a list of pointers to "Snapshot"
+func (svc *SnapshotService) GetSnapshotByName(volName string) ([]*nimbleos.Snapshot, error) {
```
rkumpf, line 76: The function name is confusing.  It should be GetSnapshotsByVolumeName.

rkumpf:
Please add ckallianpur to review the fakeservice code.

pkg/service/snapshot_service.go
vidhutsingh: This cannot be handled with this review. please don't mix.

pkg/service/volume_service.go
vidhutsingh: sure .

vidhutsingh:
> Please add ckallianpur to review the fakeservice code.

This is not fakeservice changes

pkg/service/snapshot_service.go
vidhutsingh: There is going to be next PR generated from build . you should expect the fix there .

rkumpf: I don't understand the "don't mix" comment.  If generated things are out of bounds, what is left to review?  Anyway, if you are saying the change is coming in the next PR, that is fine with me.

rkumpf:
> > Please add ckallianpur to review the fakeservice code.
> 
> This is not fakeservice changes

Was the fakeservice code included in this PR by mistake?

pkg/service/snapshot_service.go
vidhutsingh: The fix is only for specific issues. This PR doesn't contain any changes related to the snapshot.

vidhutsingh:
> > > Please add ckallianpur to review the fakeservice code.
> > 
> > 
> > This is not fakeservice changes
> 
> Was the fakeservice code included in this PR by mistake?

seems like fakeservice is included by mistake. will removed it from this PR

pkg/service/snapshot_service.go
rkumpf: Maybe the snapshot changes were also included by mistake.

rkumpf (approved):

PR #25, opened 2020-11-24
Adding tests for encrypn,pt,snapcoll,volcoll

prachiruparelia-hpe:
Signed-off-by: pruparelia <prachi.ruparelia@hpe.com>

### DIFF
--- a/test/encryption_test.go
+++ b/test/encryption_test.go
@@ -1,0 +1,95 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+
+package test
+
+import (
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+const passphrase = "passphrase-91"
+const newPassphrase = "passphrase-92"
+
+type EncryptionWorkflowSuite struct {
+	suite.Suite
+	groupService      *service.NsGroupService
+	keyManagerService *service.KeyManagerService
+	masterKeyService  *service.MasterKeyService
+}
+
+func (suite *EncryptionWorkflowSuite) SetupSuite() {
+	groupService, err := config()
+	assert.Nilf(suite.T(), err, "Unable to connect to group, err: %v", err)
+	suite.keyManagerService = groupService.GetKeyManagerService()
+	suite.masterKeyService = groupService.GetMasterKeyService()
+	suite.deleteMasterKey()
+}
+
+func (suite *EncryptionWorkflowSuite) TearDownSuite() {
+	suite.deleteMasterKey()
+}
+
+func (suite *EncryptionWorkflowSuite) deleteMasterKey() {
+	filter := &param.GetParams{}
+	key, _ := suite.masterKeyService.GetMasterKeys(filter)
+	if len(key) != 0 {
+		err := suite.masterKeyService.DeleteMasterKey(*key[0].ID)
+		assert.Nilf(suite.T(), err, "Unable to delete key, err: %v", err)
+	}
+}
+
+func (suite *EncryptionWorkflowSuite) TestCreateMasterKey() {
+	var passphrase *string = param.NewString(passphrase)
+	key := &nimbleos.MasterKey{
+		Passphrase: passphrase,
+	}
+	_, err := suite.masterKeyService.CreateMasterKey(key)
+	assert.Nilf(suite.T(), err, "Master key creation failed, err: %v", err)
+}
+
+func (suite *EncryptionWorkflowSuite) TestCreateMasterKeyDuplicate() {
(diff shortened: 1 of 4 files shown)


test/encryption_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (suite *EncryptionWorkflowSuite) SetupSuite() {
+	groupService, err := config()
+	assert.Nilf(suite.T(), err, "Unable to connect to group, err: %v", err)
+	suite.keyManagerService = groupService.GetKeyManagerService()
+	suite.masterKeyService = groupService.GetMasterKeyService()
+	suite.deleteMasterKey()
+}
+
+func (suite *EncryptionWorkflowSuite) TearDownSuite() {
```
rkumpf, line 34: Include LogoutService().

test/encryption_test.go
```diff
@@ -0,0 +1,95 @@
...
+	_, err := suite.masterKeyService.CreateMasterKey(key)
+	assert.Nilf(suite.T(), err, "Master key creation failed, err: %v", err)
+}
+
+func (suite *EncryptionWorkflowSuite) TestCreateMasterKeyDuplicate() {
+	key := &nimbleos.MasterKey{
+		Passphrase: param.NewString("passphrase1"),
+	}
+	_, err := suite.masterKeyService.CreateMasterKey(key)
+	assert.NotNil(suite.T(), err, "Master key creation when one exists should have failed")
```
rkumpf, line 62: We had a discussion earlier about state dependencies across test cases.  I don't recall the result.  Is there an intention to avoid them?

test/snapshot_collection_test.go
```diff
@@ -0,0 +1,101 @@
...
+		PeriodUnit:            nimbleos.NsPeriodUnitMinutes,
+		NumRetain:             numRetain,
+	}
+	_, err = suite.protectionScheduleService.CreateProtectionSchedule(newPS)
+	assert.Nilf(suite.T(), err, "Unable to create protection schedule", err)
+
+	filter := &param.GetParams{}
+	snapcollBefore, _ := suite.snapcollService.GetSnapshotCollections(filter)
+
+	time.Sleep(60 * time.Second)
```
rkumpf: It is unfortunate to insert this sort of delay in a test run.  I see it is probably unavoidable given the constraints of what is being tested.  However...

I still do not have an understanding of the purpose of these tests, despite asking several times.  If the intent is to test the behavior of the storage system, I'll ask why such tests belong in this project.  If the intent is to test the behavior of the Go client SDK, I'll ask why we need to verify the storage system is properly creating snapshot collections.  Wouldn't we simply verify that the protection schedule was created as requested?

prachiruparelia-hpe: I am sorry I missed replying to your question earlier. Intent is to test Go client sdk. The test case was written as such to make sure the snapshot collection is automatically triggered. But what you are saying makes sense to me. I will still confirm it with Vidhut and remove this check if its not required.

test/encryption_test.go
prachiruparelia-hpe: I started combining test cases, example in protection template "TestCreateProtectionTemplate" is also updating the protection template. But if i am going overboard its looking more like one giant test case than smaller tests. I am still

prachiruparelia-hpe: I was not sure how to achieve it besides making it one big test case. Just realized that by moving create to setup, will be able to achieve most of the test case independence. Will implement that change and push in sometime.

test/encryption_test.go
prachiruparelia-hpe: Added.

test/encryption_test.go
prachiruparelia-hpe: Moved create in Setup.

test/protection_template_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (suite *ProtectionTemplateWorkflowSuite) deleteProtectionSchedule(psName string) {
+	ps, _ := suite.protectionScheduleService.GetProtectionScheduleByName(psName)
+	if ps != nil {
+		err := suite.protectionScheduleService.DeleteProtectionSchedule(*ps.ID)
+		assert.Nilf(suite.T(), err, "Unable to delete Protection Schedule, err: %v", err)
+	}
+}
+
+func (suite *ProtectionTemplateWorkflowSuite) TestCreateProtectionTemplate() {
```
rkumpf, line 63: The function name is a bit misleading now.

test/snapshot_collection_test.go
```diff
@@ -0,0 +1,98 @@
...
+func (suite *SnapCollWorkflowSuite) deleteSnapColl(snapcollName string) {
+	snapcoll, _ := suite.snapcollService.GetSnapshotCollectionByName(snapcollName)
+	if snapcoll != nil {
+		err := suite.snapcollService.DeleteSnapshotCollection(*snapcoll.ID)
+		assert.Nilf(suite.T(), err, "Unable to delete snapshot collection, err: %v", err)
+	}
+}
+
+func (suite *SnapCollWorkflowSuite) TestCreateSnapshotCollection() {
+	// Create volume collection & associate volumes
```
rkumpf, line 57: Out of date

rkumpf (approved):

PR #27, opened 2020-11-25
Adding tests for disk, user, userpolicy, group & pool

kousalya-hpe:
Signed-off-by: Kousalya Nachimuthu <kousalya.nachimuthu@hpe.com>


test/disk_wf_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+type DiskWorkflowSuite struct {
+	suite.Suite
+	groupService *service.NsGroupService
+	diskService  *service.DiskService
+}
+
+func (suite *DiskWorkflowSuite) SetupTest() {
```
prachiruparelia-hpe: This should probably be SetupSuite().

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	fmt.Printf("sourcegroup ID %+v\n", groupID)
+
+	// Validate Group Merge
+	_, err := suite.arraygroupService.ValidateMergeGroup(groupID, sourcegroupName, sourceArrayIP, sourceArrayusername, sourceArraypassword, nil, nil)
+	assert.Nilf(suite.T(), err, "Group Merge Validate failed")
+
+	// Perform group merge
+	_, err = suite.arraygroupService.MergeGroup(groupID, sourcegroupName, sourceArrayIP, sourceArrayusername, sourceArraypassword, nil, nil, nil)
+	fmt.Printf("Wait 1 min for group merge")
+	time.Sleep(120 * time.Second)
```
prachiruparelia-hpe: Are we waiting 1 min or 2 minutes here?

test/disk_wf_test.go
kousalya-hpe: changed

test/group_pool_wf_test.go
kousalya-hpe: 2 mins, updated the comment.

test/disk_wf_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Get a disk which is in use
+	var diskID string
+	for i := 0; i < len(diskResp); i++ {
+		if *diskResp[i].State == "in use" {
+			diskID = *diskResp[i].ID
+			break
+		}
+	}
+
```
rkumpf: Add an assertion that diskID is not nil.

test/disk_wf_test.go
```diff
@@ -0,0 +1,80 @@
...
+	// Remove Disk
+	removeDisk := &nimbleos.Disk{
+		DiskOp: nimbleos.NsDiskOpRemove,
+	}
+	_, err = suite.diskService.UpdateDisk(diskID, removeDisk)
```
rkumpf: Does this cause any potential harm?  Does one need to be careful about which array is used as a target for these tests?

test/disk_wf_test.go
```diff
@@ -0,0 +1,80 @@
...
+	assert.Nilf(suite.T(), err, "Failed to Remove Disk")
+	getDiskResp, err := suite.diskService.GetDiskById(diskID)
+	assert.Nilf(suite.T(), err, "Failed to Get Disk details")
+	assert.Equal(suite.T(), nimbleos.NsDiskState("removed"), *getDiskResp.State, "In-correct User Role")
```
rkumpf: "User Role"?

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	"time"
+
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type GroupPoolWorkflowSuite struct {
```
rkumpf: Why is the same file used for both group and pool tests?

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	suite.Suite
+	groupService            *service.NsGroupService
+	arraygroupService       *service.GroupService
+	volumeService           *service.VolumeService
+	poolService             *service.PoolService
+	arrayService            *service.ArrayService
+	sourcearrayGroupService *service.GroupService
```
rkumpf: This doesn't need to be part of the suite.

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	groupID := *groupResp[0].ID
+	fmt.Printf("sourcegroup ID %+v\n", groupID)
+
+	// Validate Group Merge
+	_, err := suite.arraygroupService.ValidateMergeGroup(groupID, sourcegroupName, sourceArrayIP, sourceArrayusername, sourceArraypassword, nil, nil)
+	assert.Nilf(suite.T(), err, "Group Merge Validate failed")
+
+	// Perform group merge
+	_, err = suite.arraygroupService.MergeGroup(groupID, sourcegroupName, sourceArrayIP, sourceArrayusername, sourceArraypassword, nil, nil, nil)
+	fmt.Printf("Wait 2 mins for group merge")
```
rkumpf: Ideally the tests could be written without long delays.  I wonder if it is sufficient to check that the group merge has started and not bother waiting for it to finish.

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	filter := &param.GetParams{}
+	groupResp, err := suite.arraygroupService.GetGroups(filter)
+	groupName := *groupResp[0].Name
+	fmt.Printf("source array group name %+v\n", groupName)
+	assert.Nilf(suite.T(), err, "Get Group failed")
+	_, err = suite.arraygroupService.GetGroupByName(groupName)
+	assert.Nilf(suite.T(), err, "Get Group by name failed")
+}
+
+// Validate & Performe Group Merge
```
rkumpf: This test requires setup of multiple arrays, and it is destructive.  It seems like a lot to require just for testing the SDK.  Will there be an option to skip destructive tests?  Where will be the instructions for setting up and running the tests (and the actions required afterward if the tests are to be re-run)?

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+}
+
+// Edit Default target Scope
+func (suite *GroupPoolWorkflowSuite) TestUpdateGroupProperties() {
+	filter := &param.GetParams{}
+	groupResp, _ := suite.arraygroupService.GetGroups(filter)
+	groupID := *groupResp[0].ID
+	currentVersion := *groupResp[0].VersionCurrent
+	currentVersion = strings.Split(currentVersion, "")[0]
```
rkumpf: I think this just gets the first character from the version string.  It won't work if the major version reaches 10.

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	fmt.Printf("currentVersion: %v\n", currentVersion)
+	var newDefaultIscsiTargetScope *nimbleos.NsTargetScope
+	var newDefaultVolumeLimit int64 = 98
+
+	// Array version below 5.1 does not support group target scope
+	arrayVersion, _ := strconv.ParseFloat(currentVersion, 8)
+	var updateDefaultTargetScope *nimbleos.Group
+	if arrayVersion < 5.1 {
```
rkumpf: If we just use the first character of the version string (at line 115), version 5.2 would be considered "5" and would be less than 5.1.

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	var nonDefaulPoolID string = ""
+	for i := 0; i < poolCount; i++ {
+		if *poolResp[i].Name != "default" {
+			nonDefaulPoolName = *poolResp[i].Name
+			nonDefaulPoolID = *poolResp[i].ID
+			fmt.Printf("Non default Pool Name %+v\n", nonDefaulPoolName)
+			fmt.Printf("Non default Pool ID %+v\n", nonDefaulPoolID)
+		}
+	}
+	defaultPool := suite.GetPoolDetail("default")
```
rkumpf: Even the "default" pool can be renamed.  Is it a prerequisite for this test to have a pool named "default"?

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	// Get default pool details to get assigned arrays details
+	fmt.Printf("Starting new pool creation")
+	filter := &param.GetParams{}
+	groupResp, _ := suite.arraygroupService.GetGroups(filter)
+	defaultPool := suite.GetPoolDetail("default")
+	fmt.Printf("Pool array list %+v\n", defaultPool.ArrayList)
+	groupMemberList := groupResp[0].MemberList
+	fmt.Printf("Group memeber array list %+v\n", groupMemberList)
+	arrayCount := int(*defaultPool.ArrayCount)
+	var exisitingPoolArrayList []string
```
rkumpf: Typo *existing

test/group_pool_wf_test.go
```diff
@@ -0,0 +1,328 @@
...
+	var unassignedArrayName string
+	fmt.Printf("initiated Unassigned array name %+v\n", unassignedArrayName)
+	for i := 0; i < len(groupMemberList); i++ {
+		_, contains := contains(exisitingPoolArrayList, *groupMemberList[i])
+		if !contains {
+			unassignedArrayName = *groupMemberList[i]
+			break
+		}
+	}
+	fmt.Printf("Unassigned array name %+v\n", unassignedArrayName)
```
rkumpf: This seems like another notable prerequisite.

test/user_wf_test.go
```diff
@@ -0,0 +1,186 @@
...
+	groupService, err := config()
+	assert.Nilf(suite.T(), err, "Could not connect to array")
+	suite.groupService = groupService
+	suite.userService = groupService.GetUserService()
+	suite.userPolicyService = groupService.GetUserPolicyService()
+}
+
+// Delete User
+func (suite *UserWorkflowSuite) deleteUser(userName string) error {
+	userObj, _ := suite.userService.GetUserByName(userName)
```
rkumpf: What should happen if userObj is nil?

test/user_wf_test.go
```diff
@@ -0,0 +1,186 @@
...
+	err := suite.userService.DeleteUser(*userObj.ID)
+	return err
+}
+
+// Connect to array with different user
+func (suite *UserWorkflowSuite) connectArray(arrayIP string, username string, password string) *service.NsGroupService {
```
rkumpf: This seems like a generally useful function.  Should it be in a common place?

test/group_pool_wf_test.go
kousalya-hpe: For group & pool, I need 2 arrays to perform these tests like group merge & pool creation. So combined both the tests in same file as I will be having an additional array after group merge.

test/disk_wf_test.go
kousalya-hpe: my mistake, I'll change the message

test/disk_wf_test.go
kousalya-hpe: There won't be any harm in doing this test. We will take one 'in-use' disk slot, remove it first and add it back.

test/user_wf_test.go
kousalya-hpe: will add an assertion for userObj nil.

test/disk_wf_test.go
vidhutsingh: Add assertion

test/group_pool_wf_test.go
kousalya-hpe: Yes, assuming that the 'default' pool is present, so we could merge pool to default pool.

test/group_pool_wf_test.go
kousalya-hpe: Since these tests are added with group merge tests, we will also be performing pool merge & unassign one array from the pool. There will be a free array(unassignedArrayName) for new pool creation. So this is not a prerequisite.

test/group_pool_wf_test.go
vidhutsingh: seems ok .

test/group_pool_wf_test.go
vidhutsingh: Split will not work properly if currentVersion is 5.x . It will just get 5 . pls fix it .

test/group_pool_wf_test.go
vidhutsingh: should be okay since by default new arrays contains "default" pool .

vidhutsingh (changes requested):
please address review comments .

vidhutsingh:
Cancelling this PR and will create new PR for containervision unit test support

test/group_pool_wf_test.go
kousalya-hpe: will remove "sourcearrayGroupService"

test/user_wf_test.go
kousalya-hpe: There is already a function for array connectivity in common, but we cannot parse arguments to it. I'll see how we can merge these two functions together and will make the changes in next PR.

test/disk_wf_test.go
kousalya-hpe: added assertion for diskID

test/group_pool_wf_test.go
kousalya-hpe: corrected in code.

test/group_pool_wf_test.go
kousalya-hpe: Makes sense. Later we would be adding tests for replication as-well, which will require multi arrays. We'll come-up with some instructions &  an option which would skip running such tests. For now we can keep the name as 'group_pool_wf_xtest.go', so it will skip this test file.

vidhutsingh (approved):

kousalya-hpe:
@rkumpf are there any other comments for this PR, can I merge it ?

rkumpf:
I won't add further comments.  You have approval to merge.

PR #29, opened 2020-11-27
Implemented a sdkprovider interface for Pool and Volume object for unit testing support 

vidhutsingh:
Framework design for external thirpart application unit testing support :
sdkprovider package which contains an interface for 
1. NsGroupService
2. VolumeService 
3. PoolService 

Added fakservice package which contains fake implementations of 
1. NsGroupService methods 
2. VolumeService methods 
3. PoolService methods 

Added example/sdkprovioder.go 

The design is expandable , any new objects addition just require the addition of new object interface in sdkprovider package and implementation of corresponding methods in fakeservice package .

Testing 
go run example/sdkprovider.go 
go test -v 

### DIFF
--- a/pkg/sdkprovider/sdk_provider.go
+++ b/pkg/sdkprovider/sdk_provider.go
@@ -1,0 +1,46 @@
+package sdkprovider
+
+import (
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+)
+
+// VolumeService defines the interface to Volume related operations
+type VolumeService interface {
+	GetVolumes(params *param.GetParams) ([]*nimbleos.Volume, error)
+	CreateVolume(obj *nimbleos.Volume) (*nimbleos.Volume, error)
+	UpdateVolume(id string, obj *nimbleos.Volume) (*nimbleos.Volume, error)
+	GetVolumeById(id string) (*nimbleos.Volume, error)
+	GetVolumeByName(name string) (*nimbleos.Volume, error)
+	GetVolumeBySerialNumber(serialNumber string) (*nimbleos.Volume, error)
+	OnlineVolume(id string, force bool) (*nimbleos.Volume, error)
+	OfflineVolume(id string, force bool) (*nimbleos.Volume, error)
+	DeleteVolume(id string) error
+	AssociateVolume(volId string, volcollId string) error
+	DisassociateVolume(volId string) error
+	RestoreVolume(id string, baseSnapId string) error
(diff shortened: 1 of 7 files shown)


pkg/sdkprovider/sdk_provider.go
```diff
@@ -0,0 +1,46 @@
...
+	DeleteVolume(id string) error
+	AssociateVolume(volId string, volcollId string) error
+	DisassociateVolume(volId string) error
+	RestoreVolume(id string, baseSnapId string) error
+	MoveVolume(id string, destPoolId string, forceVvol *bool) (*nimbleos.NsVolumeListReturn, error)
+	BulkMoveVolumes(volIds []*string, destPoolId string, forceVvol *bool) (*nimbleos.NsVolumeListReturn, error)
+	AbortMoveVolume(id string) error
+	BulkSetDedupeVolumes(volIds []*string, dedupeEnabled bool) error
+	BulkSetOnlineAndOfflineVolumes(volIds []*string, online bool) error
+	DestroyVolume(id string) error
```
ckallianpur, line 27: We must implement DestroyVolume in the VolumeService.go as well otherwise it complains that VolumeService is not implementing the sdkprovider.volumeservice.

ckallianpur: github.hpe.com/nimble-dcs/golang-sdk/pkg/service
/Users/kallianc/go/pkg/mod/github.hpe.com/nimble-dcs/golang-sdk@v0.0.0-20201110084753-b00225cd1877/pkg/service/group_service_factory.go:170:20: cannot use NewVolumeService(gs) (type *VolumeService) as type sdkprovider.VolumeService in assignment:
	*VolumeService does not implement sdkprovider.VolumeService (missing DestroyVolume method)

This is the exact error

ckallianpur (commented):
Looks good to me. If we use it as it is it will fail since the stats method is not implemented. But as discussed, we can enhance the sdk before using these set of changes

pkg/sdkprovider/sdk_provider.go
vidhutsingh: You are checking in private GitHub where it doesnt exist . You should check in public git repo . I have already validated this and it works. 
In private repo this code is not checked in yet . it will be done after public git PR is approved.

ckallianpur: thanks. I see DestroyVolume. Since array-exporter was using private github, I copied all changes to that.

ckallianpur (approved):

vidhutsingh:
Merge is still blocked even after Chaitra approval, I think one more reviewer approval is needed. @roger, can you approve this change?

rkumpf (approved):

vidhutsingh:
thanks roger .

PR #31, opened 2020-12-02
CreatePool method was missing from interface definition, fixed it

vidhutsingh:
Fix :
Added CreatePool method signature to the interface definition 


rkumpf (approved):

vidhutsingh:
That was fast , thanks roger .

PR #36, opened 2020-12-04
SA-202, return full http response error message

vidhutsingh:

### DIFF
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -393,7 +393,7 @@ func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, er
 			}
 		}
 		if len(jobId) == 0 {
-			return nil, fmt.Errorf("http response error: status (202), failed to get the job id")
+			return nil, fmt.Errorf("http response error: status (202), failed to get the job id, message: %v",errResp)
 		}
 
 		id, err := waitForJobResult(jobId, client)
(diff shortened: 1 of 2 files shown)


pkg/client/group_mgmt_client.go
```diff
@@ -393,7 +393,7 @@ func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, er
 			}
 		}
 		if len(jobId) == 0 {
-			return nil, fmt.Errorf("http response error: status (202), failed to get the job id")
+			return nil, fmt.Errorf("http response error: status (202), failed to get the job id, message: %v",errResp)
```
rkumpf: I suggest changing "failed to get the job id" to "but no job ID returned".  Also change "message" to "messages" for consistency with the other errors.

pkg/client/group_mgmt_client.go
```diff
@@ -393,7 +393,7 @@ func processAsyncResponse(client *GroupMgmtClient, body []byte) (interface{}, er
 			}
 		}
 		if len(jobId) == 0 {
```
rkumpf, line 395: At line 375, errResp is constructed with no separator between the messages.  I'm curious to see how it looks with multiple messages.

vidhutsingh: if you notice there is a default separator in the message itself and we don't want to add a custom separator.

pkg/client/group_mgmt_client.go
vidhutsingh: sure

pkg/client/group_mgmt_client.go
rkumpf: I can't tell that there is a separator in the message, but I'll take your word for it.

rkumpf (approved):

vidhutsingh:
thanks rkumpf

PR #40, opened 2020-12-04
Rep Partner, FC and fixes

prachiruparelia-hpe:
Signed-off-by: pruparelia <prachi.ruparelia@hpe.com>

Changes in this PR:
- Support to pass in array details from command line
- fcenabled, iscsienabled and array version check support
- Fibrechannel and replication partner test cases
- FC initiator test cases

### DIFF
--- a/test/common.go
+++ b/test/common.go
@@ -8,6 +8,8 @@ import (
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/sdkprovider"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"strconv"
+	"strings"
 )
 
 const defaultVolumeName = "DefaultVolumeTest"
@@ -20,10 +22,14 @@ var arrayUsername = flag.String("arrayUsername", "xxx", "Array Username")
 
 var arrayPassword = flag.String("arrayPassword", "xxx", "Array Password")
 
+var downstreamArrayIP = flag.String("downstream", "", "Array IP to be used as downstream")
+
 // Required for group merge test
-const sourceArrayIP = "1.1.1.2"
-const sourceArrayusername = "xxx"
-const sourceArraypassword = "xxx"
+var sourceArrayIP = flag.String("sourceArrayIP", "1.1.1.2", "Source array IP for group tests")
+
+var sourceArrayusername = flag.String("sourceArrayusername", "xxx", "Source array username")
+
+var sourceArraypassword = flag.String("sourceArraypassword", "xxx", "Source array password")
 
 func config() (*service.NsGroupService, error) {
 	groupService, err := service.NewNsGroupService(*arrayIP, *arrayUsername, *arrayPassword, "v1", true)
@@ -87,4 +93,28 @@ func deleteDefaultVolColl(volcollService *service.VolumeCollectionService) error
 		return err
 	}
 	return volcollService.DeleteVolumeCollection(*volcoll.ID)
+}
+
+func isFCEnabled(arrayGroupService *service.GroupService) bool {
+	filter := &param.GetParams{}
+	groups, _ := arrayGroupService.GetGroups(filter)
+	group := groups[0]
+	return *group.FcEnabled
+}
+
+func isIscsiEnabled(arrayGroupService *service.GroupService) bool {
+	filter := &param.GetParams{}
+	groups, _ := arrayGroupService.GetGroups(filter)
+	group := groups[0]
+	return *group.IscsiEnabled
+}
+
+func getArrayVersion(groupService *service.GroupService) float64 {
+	filter := &param.GetParams{}
+	groupResp, _ := groupService.GetGroups(filter)
+	currentVersion := *groupResp[0].VersionCurrent
+	currentVersion = strings.Split(strings.TrimSpace(currentVersion), "-")[0]
+	currentVersionSplitlist := strings.Split(currentVersion, ".")[:2]
+	arrayVersion, _ := strconv.ParseFloat(strings.Join(currentVersionSplitlist, "."), 8)
+	return arrayVersion
 }
--- a/test/common.go
+++ b/test/common.go
@@ -16,16 +16,16 @@ const defaultVolumeName = "DefaultVolumeTest"
 const defaultInitiatorGrpName = "DefaultInitiatorgrpTest"
 const defaultVolCollName = "DefaultVolCollTest"
 
-var arrayIP = flag.String("arrayIP", "1.1.1.1", "Array IP")
+var arrayIP = flag.String("arrayIP", "", "Array IP")
 
 var arrayUsername = flag.String("arrayUsername", "xxx", "Array Username")
 
 var arrayPassword = flag.String("arrayPassword", "xxx", "Array Password")
 
-var downstreamArrayIP = flag.String("downstream", "", "Array IP to be used as downstream")
+var downstreamArrayIP = flag.String("downstream", "", "IP address of an array to be used as a downstream replica. Pass this IP to execute replication partner test cases.")
 
 // Required for group merge test
-var sourceArrayIP = flag.String("sourceArrayIP", "1.1.1.2", "Source array IP for group tests")
+var sourceArrayIP = flag.String("sourceArrayIP", "", "IP address of source array, used for group and merge test cases.")
 
 var sourceArrayusername = flag.String("sourceArrayusername", "xxx", "Source array username")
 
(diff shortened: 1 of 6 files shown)


test/common.go
```diff
@@ -20,10 +22,14 @@ var arrayUsername = flag.String("arrayUsername", "xxx", "Array Username")
 
 var arrayPassword = flag.String("arrayPassword", "xxx", "Array Password")
 
+var downstreamArrayIP = flag.String("downstream", "", "Array IP to be used as downstream")
+
 // Required for group merge test
-const sourceArrayIP = "1.1.1.2"
-const sourceArrayusername = "xxx"
-const sourceArraypassword = "xxx"
+var sourceArrayIP = flag.String("sourceArrayIP", "1.1.1.2", "Source array IP for group tests")
```
rkumpf: The default value is not useful.  Just leave it as an empty string.  Same for the other flags.

test/common.go
```diff
@@ -88,3 +94,27 @@ func deleteDefaultVolColl(volcollService *service.VolumeCollectionService) error
...
+	return *group.IscsiEnabled
+}
+
+func getArrayVersion(groupService *service.GroupService) float64 {
+	filter := &param.GetParams{}
+	groupResp, _ := groupService.GetGroups(filter)
+	currentVersion := *groupResp[0].VersionCurrent
+	currentVersion = strings.Split(strings.TrimSpace(currentVersion), "-")[0]
+	currentVersionSplitlist := strings.Split(currentVersion, ".")[:2]
+	arrayVersion, _ := strconv.ParseFloat(strings.Join(currentVersionSplitlist, "."), 8)
```
rkumpf, line 124: Why only 8 bits of precision?  I suggest 64, to minimize roundoff errors.  Floating point numbers are difficult to compare precisely.

test/common.go
```diff
@@ -88,3 +94,27 @@ func deleteDefaultVolColl(volcollService *service.VolumeCollectionService) error
 	}
 	return volcollService.DeleteVolumeCollection(*volcoll.ID)
 }
+
+func isFCEnabled(arrayGroupService *service.GroupService) bool {
+	filter := &param.GetParams{}
+	groups, _ := arrayGroupService.GetGroups(filter)
+	group := groups[0]
+	return *group.FcEnabled
```
rkumpf: Note that a nil dereference is possible, causing a panic.  Maybe that is okay for this test code.

test/common.go
```diff
@@ -20,10 +22,14 @@ var arrayUsername = flag.String("arrayUsername", "xxx", "Array Username")
...
 
+var downstreamArrayIP = flag.String("downstream", "", "Array IP to be used as downstream")
+
 // Required for group merge test
-const sourceArrayIP = "1.1.1.2"
-const sourceArrayusername = "xxx"
-const sourceArraypassword = "xxx"
+var sourceArrayIP = flag.String("sourceArrayIP", "1.1.1.2", "Source array IP for group tests")
+
+var sourceArrayusername = flag.String("sourceArrayusername", "xxx", "Source array username")
```
rkumpf, line 30: The flag names are getting long.  It's not a problem, but it might be inconvenient.

Also, the 'u' in 'username' should be upper case.

test/common.go
```diff
@@ -20,10 +22,14 @@ var arrayUsername = flag.String("arrayUsername", "xxx", "Array Username")
...
+
+var sourceArraypassword = flag.String("sourceArraypassword", "xxx", "Source array password")
```
rkumpf, line 32: Please explain what a source array is in the description.

test/common.go
```diff
@@ -20,10 +22,14 @@ var arrayUsername = flag.String("arrayUsername", "xxx", "Array Username")
 
 var arrayPassword = flag.String("arrayPassword", "xxx", "Array Password")
 
+var downstreamArrayIP = flag.String("downstream", "", "Array IP to be used as downstream")
```
rkumpf: Please explain what a 'downstream' is.

test/common.go
```diff
@@ -88,3 +94,27 @@ func deleteDefaultVolColl(volcollService *service.VolumeCollectionService) error
...
+
+func isFCEnabled(arrayGroupService *service.GroupService) bool {
+	filter := &param.GetParams{}
+	groups, _ := arrayGroupService.GetGroups(filter)
+	group := groups[0]
+	return *group.FcEnabled
+}
+
+func isIscsiEnabled(arrayGroupService *service.GroupService) bool {
+	filter := &param.GetParams{}
```
rkumpf, line 109: We seem to make a lot of repetitive calls to get the group settings.  If we don't consolidate them, we could at least specify a Fields parameter to limit the response to just the fields we need.  It seems to be just a single field in many cases.

test/initiator_group_test.go
```diff
@@ -19,6 +19,9 @@ type IGWorkflowSuite struct {
 	groupService        *service.NsGroupService
 	initiatorGrpService *service.InitiatorGroupService
 	volumeService       sdkprovider.VolumeService
+	arrayGroupService   *service.GroupService
+	fcEnabled           bool
```
rkumpf: Unused?

test/initiator_group_test.go
```diff
@@ -66,19 +71,33 @@ func (suite *IGWorkflowSuite) TestCreateIG() {
 	assert.Nilf(suite.T(), err, "Unable to create initiator group, err: %v", err)
 	ig, _ := suite.initiatorGrpService.GetInitiatorGroupByName(initiatorGroupName)
 	assert.Equal(suite.T(), *desc, *ig.Description, "Initiator group creation does not have expected description.")
-}
+	var newDesc *string = param.NewString("Workflow tests for initiator group")
+	updateIG := &nimbleos.InitiatorGroup{
+		Description: newDesc,
+	}
+	if ig != nil {
```
rkumpf: Remove this condition.  If ig is nil, we already panicked at line 73.

test/initiator_group_test.go
```diff
@@ -94,9 +113,49 @@ func (suite *IGWorkflowSuite) TestCreateIGIscsiInitiators() {
 	}
 	_, err := suite.initiatorGrpService.CreateInitiatorGroup(newIG)
 	assert.Nilf(suite.T(), err, "Unable to create initiator group, err: %v", err)
+	// Update initiator group
+	var ipAdd *string = param.NewString("10.1.0.0")
```
rkumpf: This could be simplified to:

   ipAdd := "10.1.0.0"

And then below, change ipAdd to &ipAdd and change *ipAdd to ipAdd.

test/replication_partner_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/sdkprovider"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+const repPartner = "ReplcationPartnerTest"
+const repPassphrase = "Nim123Boli"
```
rkumpf, line 15: Is this the best choice?

test/common.go
prachiruparelia-hpe: I couldnt change this value since Kousalya has a PR out which uses lower case username.
@kousalya-hpe can you please update this change in your PR too?

test/common.go
vidhutsingh: Replication direction from Upstream Array to Downstream Array.

test/initiator_group_test.go
vidhutsingh: That should fine at least here since param.NewString is being leveraged to get the pointer to a string in one line. I am glad to see param package is being used which I will encourage while writing SDK-related code. thought ?

rkumpf: Line 120 and others show appropriate usage of param.NewString.  No need to overuse it.

test/common.go
rkumpf: I meant to explain it in the option description.  For example: "IP address of an array to be used as a downstream replica in tests that configure replication".

If this option is not always required, please specify when it must be provided.  Same for the other options.

test/common.go
vidhutsingh: Good suggestion. please file an enhancement bug for it. In the interest of time let's keep it .

vidhutsingh:
thanks for addressing review comment.

vidhutsingh (approved):

vidhutsingh:
@rkumpf , do you have any other concerns or shall we merge this change?

rkumpf (commented):
I think there are a few easy improvements that could be made, but the patch seems fine to merge.

vidhutsingh:
> I think there are a few easy improvements that could be made, but the patch seems fine to merge.

Thanks @rkumpf . 
@prachi, please file a separate enhancement bug to track easy improvements suggestion.

PR #42, opened 2021-01-07
Bug fix SA-203 - AvgStatsLast5mins is not populating correctly as it always display nil 

vidhutsingh:
Bug SA-203 
The volume average stats for last 5mins always display nil and doesnt poplate it correctly.

Fix :

v1/nimbleos/volume.go , file contains the definition of volume struct with supported fields and json tags. The json tag for the field AvgStatsLast5mins has a typo . It has "avg_stats_last5mins" instead of "avg_stats_last_5mins" which causes to return default value nil for AvgStatsLast5mins. Made the tag name changes to fix the issue . 

Unit test:
Ran the unit test to validate it correctly populate the filed values. 

### DIFF
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -417,7 +417,7 @@ func waitForJobResult(jobId string, client *GroupMgmtClient) (interface{}, error
 
 			job, err := client.GetJobObjectSet().GetObject(jobId)
 			if err != nil {
-				fmt.Printf("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+				fmt.Printf("Warning : failed to get %s jobId info, err : %s\n", jobId, err.Error())
 			} else {
 				var objectId = *job.ObjectId
 				if string(*job.State) == string(*nimbleos.NsJobStatusDone) {
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -417,7 +417,7 @@ func waitForJobResult(jobId string, client *GroupMgmtClient) (interface{}, error
 
 			job, err := client.GetJobObjectSet().GetObject(jobId)
 			if err != nil {
-				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+				fmt.Printf("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
 			} else {
 				var objectId = *job.ObjectId
 				if string(*job.State) == string(*nimbleos.NsJobStatusDone) {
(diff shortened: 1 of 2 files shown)


pkg/client/group_mgmt_client.go
```diff
@@ -417,7 +417,7 @@ func waitForJobResult(jobId string, client *GroupMgmtClient) (interface{}, error
 
 			job, err := client.GetJobObjectSet().GetObject(jobId)
 			if err != nil {
-				fmt.Println("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
+				fmt.Printf("Warning : failed to %s jobId info, err : %s", jobId, err.Error())
```
rkumpf: Add '\n' to the end of the string.

rkumpf (approved):
Add a unit test?

vidhutsingh:
> Add a unit test?

It was a typo in the field name which caused this issue. I think we should not add unit test case of field name typo however i am open to suggestion

pkg/client/group_mgmt_client.go
vidhutsingh: Sure thanks

rkumpf:
> > Add a unit test?
> 
> It was a typo in the field name which caused this issue. I think we should not add unit test case of field name typo however i am open to suggestion

Was it actually a typo, as in a person mistyping it?  I might have guessed it came out wrong from the generator, which could indicate a potential class of problems that might recur.

Previously we discussed the absence of unit tests in this project, and George indicated they are needed.  The SDK behavior should be verified, and regressions identified, without requiring a connection to another network device.

For this issue, you were unable to test a proposed fix because you lacked access to an array that returned non-zero data.  So you called on me to verify it.  I wrote a unit test to demonstrate that access to an array was not required (and because I also lack access to such an array).  I do not understand why it would be considered preferable not to have unit tests.

vidhutsingh:
> > > Add a unit test?
> > 
> > 
> > It was a typo in the field name which caused this issue. I think we should not add unit test case of field name typo however i am open to suggestion
> 
> Was it actually a typo, as in a person mistyping it? I might have guessed it came out wrong from the generator, which could indicate a potential class of problems that might recur.
> 
> Previously we discussed the absence of unit tests in this project, and George indicated they are needed. The SDK behavior should be verified, and regressions identified, without requiring a connection to another network device.
> 
> For this issue, you were unable to test a proposed fix because you lacked access to an array that returned non-zero data. So you called on me to verify it. I wrote a unit test to demonstrate that access to an array was not required (and because I also lack access to such an array). I do not understand why it would be considered preferable not to have unit tests.

Does it really matter who does it? typographical mistake remains the same. There was a bug in the Generator which caused this typo and instead of generating avg_stats_last_5mins it generated avg_stats_last5mins.
Ideally, the unit test for this should go to the generator not in the SDK. since it is a Generator bug. 

There is no absence of unit test, regression test, and integration test in this project. you are assuming that you should be able to run the unit test without Array that was never a requirement from start, that is nice to have a suggestion not mandatory. We are already running a regression test with every build. please don't assume.  
1. volume_service_test.go - contains the unit tests . 
2. pkg/test - contains functional and integration tests . 

Appreciate your help, I didn't know even you don't have the actual Array setup, if you could have said so I would figure it out.

PR #46, opened 2021-04-29
SA-207, Go SDK gets stuck in indefinite login retry if array password is changed . Applicable only for long running job. 

vidhutsingh:
The issue will occur under following conditions:
Long running application.
User login to the Nimble Array and change the password while application is still running.
 
Although, it is a very corner case but still serious bug with SDK since it get stuck in indefinite retry login attempt and never returns to app . Thanks for file the defect.
 
Quick workaround would be to  restart the application however this may not be acceptable solution.
 
RCA :
group_mgmt_client.go :
 
AddRetryCondition() is not working properly as it is recursively calling
sessionToken, err := groupMgmtClient.login(username, password) since the password has changed on Array and the SDK still uses the old username , password so it never returns to app.  
 
Code fix:
I have added a new variable “loginRetry” in AddRetryCondition() which keep track of no of login attempt , if it exhaust the default 5
It returns to caller. Also inside the AddRetryCondition() , we should create a new instance of groupMgmtClient and replace the new session_token with old session token with original groupMgmtClient.
 
The code fix changes has worked and I will send the code for review today by eod.
 
### DIFF
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -17,7 +17,7 @@ import (
 const (
 	groupURIFmt     = "https://%s:5392/%s"
 	clientTimeout   = time.Second * 60 // 1 Minute
-	maxLoginRetries = 5
+	maxLoginRetries = 1
 	jobTimeout      = time.Second * 300 // 5 Minute
 	jobPollInterval = 5 * time.Second   // Second
 	smAsyncJobId    = "SM_async_job_id"
@@ -89,34 +89,31 @@ func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool
 	}
 	groupMgmtClient.SessionToken = sessionToken
 
-	// Login retry counter
-	loginRetry := 0
 	// Add retryCondition
+	// This flag is set during a relogin attempt to prevent retry recursion
+	reloginInProgress := false
+
+	// Add a retry condition to perform a relogin if the session has expired
 	groupMgmtClient.Client.
 		AddRetryCondition(
 			func(resp *resty.Response, err error) bool {
-
-				if err == nil && resp.StatusCode() == 401 {
-					// login again and refreh the session token
-					loginRetry++
-					if loginRetry == maxLoginRetries {
-						// exhausted the login attempt. backoff now .
+				// Attempt relogin on an authorization error if relogin is not already in progress
+				if err == nil && resp.StatusCode() == 401 && !reloginInProgress {
+					reloginInProgress = true
+					sessionToken, err = groupMgmtClient.login(username, password)
+					reloginInProgress = false
+					if err != nil {
 						return false
 					}
-					newGroupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs)
-					sessionToken, err = newGroupMgmtClient.login(username, password)
-					if err != nil {
-						// returning true would retry the login again with new instance of group mgmt client
-						return true
-					}
+
 					// replace the original client session token with new session token
 					groupMgmtClient.SessionToken = sessionToken
 					resp.Request.SetHeader("X-Auth-Token", groupMgmtClient.SessionToken)
-					return false
+					return true
 				}
 				return false
 			},
-		).SetRetryCount(1)
+		).SetRetryCount(maxLoginRetries)
 	return groupMgmtClient, nil
 }
 
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -61,21 +61,28 @@ type Argument struct {
 	JobId string `json:"job_id,omitempty"`
 }
 
-// NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
-	// Create GroupMgmt Client
-	client := resty.New()
-	client.SetTLSClientConfig(&tls.Config{
+func newGroupMgmtClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) *GroupMgmtClient {
+	// Get new resty()
+	restyClient := resty.New()
+	restyClient.SetTLSClientConfig(&tls.Config{
 		InsecureSkipVerify: true,
 	})
+
+	// Create GroupMgmt Client
 	groupMgmtClient := &GroupMgmtClient{
 		URL:       fmt.Sprintf("https://%s:5392/%s", ipAddress, apiVersion),
-		Client:    client,
+		Client:    restyClient,
 		WaitOnJob: waitOnJobs,
 		Username:  username,
 		Password:  password,
 	}
+	return groupMgmtClient
+}
 
+// NewClient instantiates a new client to communicate with the Nimble group
+func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
+	// Get resty client
+	groupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs)
 	// Get session token
 	sessionToken, err := groupMgmtClient.login(username, password)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/client/group_mgmt_client.go
```diff
@@ -61,46 +61,63 @@ type Argument struct {
...
-		SetRetryCount(maxOpsRetries).
 		AddRetryCondition(
 			func(resp *resty.Response, err error) bool {
+
 				if err == nil && resp.StatusCode() == 401 {
 					// login again and refreh the session token
-					sessionToken, err = groupMgmtClient.login(username, password)
-					if err != nil {
+					loginRetry++
+					if loginRetry == maxLoginRetries {
```
rkumpf: This seems like it will limit the number of times a long-running client can recover from a session expiration, which would be problematic.

pkg/client/group_mgmt_client.go
```diff
@@ -61,46 +61,63 @@ type Argument struct {
...
+						// exhausted the login attempt. backoff now .
 						return false
 					}
+					newGroupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs)
```
rkumpf: If I understand correctly, the reason to create a new client here is so we can perform a login without a retry condition.  Could we instead use the existing client and use a flag to indicate that we are performing a re-login, so the retry condition knows not to try another re-login?  It seems better to avoid creating extra client connections.

pkg/client/group_mgmt_client.go
```diff
@@ -61,46 +61,63 @@ type Argument struct {
...
+					// replace the original client session token with new session token
 					groupMgmtClient.SessionToken = sessionToken
 					resp.Request.SetHeader("X-Auth-Token", groupMgmtClient.SessionToken)
-					return true
+					return false
 				}
 				return false
 			},
-		)
+		).SetRetryCount(1)
```
rkumpf: The maxOpsRetries constant is no longer used and can be removed.

pkg/client/group_mgmt_client.go
```diff
@@ -61,46 +61,63 @@ type Argument struct {
...
-					if err != nil {
+					loginRetry++
+					if loginRetry == maxLoginRetries {
+						// exhausted the login attempt. backoff now .
 						return false
 					}
+					newGroupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs)
+					sessionToken, err = newGroupMgmtClient.login(username, password)
+					if err != nil {
+						// returning true would retry the login again with new instance of group mgmt client
```
rkumpf: Why is another retry desirable?

pkg/client/group_mgmt_client.go
vidhutsingh: We don't want to have an indefinite login retry attempt otherwise we would get into a hung state. The default value for maxLoginRetries is eq 5 . we can bump it up if required. However, this is a decent retry attempt.

pkg/client/group_mgmt_client.go
vidhutsingh: Hey No, the groupMgmtClient.login(username, password)  function is getting called recursively on the same groupMgmtClient instance and it never returns if the password is invalid which is happening in this case. We need to have a new instance of groupMgmtClient to avoid the recursive call. Hope this clarifies.

pkg/client/group_mgmt_client.go
vidhutsingh: maxOpsRetries can be removed if it is not used .

pkg/client/group_mgmt_client.go
vidhutsingh: We want to retry until "if loginRetry == maxLoginRetries", so even if we get err via login attempt we would still retry. 
So retry backoff only when it hit loginRetry == maxLoginRetries condition.

pkg/client/group_mgmt_client.go
rkumpf: Yes, like I said, a new client is used to avoid retry recursion.  However, I don't think we need to use a new client to do that.  I suggest reusing the same client like this:

	// This flag is set during a relogin attempt to prevent retry recursion
	reloginInProgress := false

	// Add a retry condition to perform a relogin if the session has expired
	groupMgmtClient.Client.
		AddRetryCondition(
			func(resp *resty.Response, err error) bool {
				// Attempt relogin on an authorization error if relogin is not already in progress
				if err == nil && resp.StatusCode() == 401 && !reloginInProgress {
					reloginInProgress = true
					sessionToken, err = groupMgmtClient.login(username, password)
					reloginInProgress = false
					if err != nil {
						return false
					}

					// replace the original client session token with new session token
					groupMgmtClient.SessionToken = sessionToken
					resp.Request.SetHeader("X-Auth-Token", groupMgmtClient.SessionToken)
					return true
				}
				return false
			},
		).SetRetryCount(1)

pkg/client/group_mgmt_client.go
rkumpf: I agree that we should not have unlimited consecutive retries.  My concern with this patch is that once we go through a number of cycles of session expiration and refresh, we'll hit a limit and the next session expiration will be unrecoverable.

pkg/client/group_mgmt_client.go
rkumpf: I think the only time we want to do a retry is if the session token might have expired.  That doesn't seem to be the case here.

pkg/client/group_mgmt_client.go
```diff
@@ -61,46 +60,63 @@ type Argument struct {
...
+						return true
+					}
+					// replace the original client session token with new session token
 					groupMgmtClient.SessionToken = sessionToken
 					resp.Request.SetHeader("X-Auth-Token", groupMgmtClient.SessionToken)
-					return true
+					return false
```
rkumpf: Don't we need to return true here so that the original request will be retried?

pkg/client/group_mgmt_client.go
vidhutsingh: You should handle this in your app . If login attempt is exhausted and sdk returned error , you should retry in your app with updated credentials.

pkg/client/group_mgmt_client.go
vidhutsingh: Your proposed logic will go in indefinite recursion and it will never reach this line "reloginInProgress = false". 
Even earlier implementation has similar issue that's the reason I have made the code change.

pkg/client/group_mgmt_client.go
vidhutsingh: do you know we get same error code in both the  cases i. e expired token and invalid credentials

pkg/client/group_mgmt_client.go
vidhutsingh: No.

pkg/client/group_mgmt_client.go
vidhutsingh: If this is still not clear to you, I can set up a call and explain. thanks

pkg/client/group_mgmt_client.go
rkumpf: The application should definitely own the retry if the credentials have changed.  Forcing the application to retry on a session expiration is defeats the purpose of the automatic token refresh logic.

pkg/client/group_mgmt_client.go
rkumpf: Is status code 401 returned in both cases?  Yes, it must be.  The expired token case already depends on a 401 status.  The invalid credentials case would not have encountered retry recursion if it did not return status code 401.

pkg/client/group_mgmt_client.go
rkumpf: Care to explain?

pkg/client/group_mgmt_client.go
rkumpf: It is still not clear to me.  I thought the "!reloginInProgress" condition before the relogin attempt would prevent recursion.

FWIW, George's preference seems to be to avoid creating another client connection.  https://github.com/hpe-storage/nimble-golang-sdk/pull/12#discussion_r498813718

I can meet now if you want to set up a call.

vidhutsingh: @rkumpf , made the suggested changes and seems to be working . Sorry i missed "reloginInProgress = true" in your porposed code . thanks

rkumpf (approved):

PR #48, opened 2021-05-07
Auto generated Go sdk files 

vidhutsingh:
Signed-off-by: vidhutsingh <vidhut-kumar.singh@hpe.com>

### DIFF
--- a/pkg/client/v1/nimbleos/ns_j_w_k_algorithm.go
+++ b/pkg/client/v1/nimbleos/ns_j_w_k_algorithm.go
@@ -1,0 +1,21 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+package nimbleos
+
+// Golang package for NsJWKAlgorithm Enum.
+
+type NsJWKAlgorithm string
+
+const (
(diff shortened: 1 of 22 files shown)


pkg/client/v1/nimbleos/ns_j_w_k_algorithm.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
+package nimbleos
+
+// Golang package for NsJWKAlgorithm Enum.
```
rkumpf, line 4: It seems appropriate to add JWK to the list of acronyms with special treatment when converting between symbol style, so that this file name is changed from ns_j_w_k_algorithm.go to ns_jwk_algorithm.go. Maybe something similar for ns_j_w_key.go too.

rkumpf (approved):

pkg/client/v1/nimbleos/ns_j_w_k_algorithm.go
vidhutsingh: Sure, I will file a separate bug for this since it has to be done in the Go Generator code.

PR #50, opened 2021-06-01
Golang dashboard changes for all testsuites

prasannakumari-sundaraeasan:
Signed-off-by: prasannakumari-sundaresaesan <prasannakumari.sundaresaesan@hpe.com>

### DIFF
--- a/test/access_control_records_test.go
+++ b/test/access_control_records_test.go
@@ -3,12 +3,13 @@
 package test
 
 import (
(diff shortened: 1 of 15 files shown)


vidhutsingh (commented):
Why we can't use this https://github.com/vakenbolt/go-test-report to captured the test report in the dashboard? Instead of building own logic of doing the dashboard reporting.

PR #53, opened 2021-07-14
Generated changes from SM

IrennaLumbuun:
Change endpoints based on tenantAware parameter
Other files are generated by SM


pkg/client/client_credentials_object_set.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
```
rkumpf: 2021?

pkg/client/group_mgmt_client.go
```diff
@@ -31,6 +32,7 @@ type GroupMgmtClient struct {
 	WaitOnJob    bool
 	Username     string
 	Password     string
+	TenantAware  bool
```
rkumpf: Would "IsTenant" be more appropriate?  This patch seems to make the SDK tenant-aware, but this settings seems to indicate whether we are creating a client instance on behalf of a tenant user.  Same thing at line 65, etc.

pkg/client/v1/nimbleos/client_credential.go
```diff
@@ -0,0 +1,42 @@
...
+
+package nimbleos
+
+
+// ClientCredential - Credential that this device will trust.
+// Export ClientCredentialFields for advance operations like search filter etc.
+var ClientCredentialFields *ClientCredential
+
+func init(){
+ IDfield:= "id"
```
rkumpf: Why single-character indentation?

pkg/client/v1/nimbleos/client_credential.go
```diff
@@ -0,0 +1,42 @@
...
+  ClientId:     &ClientIdfield,
+  Secret:       &Secretfield,
+  Description:  &Descriptionfield,
+ }
+}
+
+
+type ClientCredential struct {
+ // ID - Identifier for the client credentials record.
+  ID *string `json:"id,omitempty"`
```
rkumpf: Indentation is inconsistent between the fields and the comments.

pkg/client/v1/nimbleos/client_credential.go
vidhutsingh: please use "go fmt" , formatting is off.

pkg/client/v1/nimbleos/client_credential.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 Hewlett Packard Enterprise Development LP
```
vidhutsingh: Please change 2020 =>2021

vidhutsingh:
Can we also add a unit test around tenant awareness?

vidhutsingh:
Please fix DCO error . you can use "git commit -s "

pkg/client/client_credentials_object_set.go
```diff
@@ -1,12 +1,23 @@
+<<<<<<< HEAD
 // Copyright 2020 Hewlett Packard Enterprise Development LP
+=======
+// Copyright 2021 Hewlett Packard Enterprise Development LP
+>>>>>>> 87138d62 (Add autogenerated files)
```
raunakkumar: Looks like an issue here. git rebase/merge seems to have a conflict

IrennaLumbuun: Fixed! thanks!

pkg/client/group_mgmt_client.go
```diff
@@ -60,28 +62,34 @@ type Argument struct {
...
 		Username:  username,
 		Password:  password,
+		isTenant:  isTenant,
 	}
 	return groupMgmtClient
 }
 
 // NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
+func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs, tenantAware bool) (*GroupMgmtClient, error) {
```
rkumpf: Should be isTenant.

pkg/client/access_control_records_object_set.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Hewlett Packard Enterprise Development LP
+// Copyright 2021 Hewlett Packard Enterprise Development LP
```
rkumpf: I'm not confident in my understanding of copyright law, but I think it should be "2020-2021".

examples/fields.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func main() {
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
+	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true, false)
```
rkumpf: If we used the functional options pattern for this constructor, we wouldn't impact existing consumers each time a constructor option is added.  It can be disruptive for a consumer when simply updating a dependency version to have to figure out why compilation is broken and how to fix it.  If it is not reasonable to address it as part of this change, perhaps it could be taken up as a separate enhancement?

pkg/client/v1/nimbleos/ns_o_t_p_type.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2021 Hewlett Packard Enterprise Development LP
+package nimbleos
+
+// Golang package for NsOTPType Enum.
+
+type NsOTPType string
```
rkumpf: I believe the generator has an acronym list to improve symbol style conversions.  Adding "OTP" to the list would improve this file name, changing it from ns_o_t_p_type.go to ns_otp_type.go.

pkg/client/v1/nimbleos/volume.go
```diff
@@ -268,7 +268,7 @@ type Volume struct {
 	// PreFilter - Pre-filtering criteria.
 	PreFilter *string `json:"pre_filter,omitempty"`
 	// AvgStatsLast5mins - Average statistics in last 5 minutes.
-	AvgStatsLast5mins *NsAverageStats `json:"avg_stats_last_5mins,omitempty"`
+	AvgStatsLast5mins *NsAverageStats `json:"avg_stats_last5mins,omitempty"`
```
rkumpf: Oh, ouch.  This defect keeps recurring.  It should have caused a test failure this time, at least.

pkg/service/group_service_factory.go
```diff
@@ -16,63 +16,64 @@ type NsGroupService struct {
 	client *client.GroupMgmtClient
 
 	// Declare all the supported services
-	accessControlRecordService        *AccessControlRecordService
```
rkumpf, line 19: I wonder why the order changed.  I feel like this happened before, too.  It makes it difficult to tell if there are any impactful changes here.

examples/filters.go
```diff
@@ -12,7 +12,7 @@ import (
 func main() {
 
 	// login to Array, get groupService instance
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
+	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false, false)
```
rkumpf: Remove the extra parameter through out these examples.  (I see 5 instances.)

pkg/client/v1/nimbleos/volume.go
rkumpf: What is the resolution here?  It looks like the missing underscore defect is still present.  A fix is needed in the generator.  Until that happens, a manual fix in the generated code will be required after each run.  (Probably this was fixed before another generator run was done for the otp file name change?  Manual changes are a pain.)

pkg/client/group_mgmt_client.go
```diff
@@ -69,19 +79,29 @@ func newGroupMgmtClient(ipAddress, username, password, apiVersion string, waitOn
...
+
 	return groupMgmtClient
 }
 
 // NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
+func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool, clientOpts ...ClientOption) (*GroupMgmtClient, error) {
 	// Get resty client
-	groupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs)
+	groupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs, clientOpts...)
```
rkumpf: Thank you for making this change to avoid impacts to existing consumers.  It has me thinking, though....

We're creating a mixture of two factory styles.  The convention for functional options is to use With*() functions rather than Is*() functions.  WithTenant() isn't a great choice because acts more like a modifier for the user name rather than its own thing.  The mixture of styles doesn't let us combine the tenant concept with the user concept.

Also, the waitOnJobs flag shouldn't always need to be specified.  It is a ripe candidate for a functional option.  And apiVersion is always "v1" for the forseeable future.

So I'm wondering if it would be better to essentially deprecate the NewClient() function, leaving it unmodified without tenant support, and introduce a new factory that only uses functional options.

There is an issue of what to call the new factory function, since overloading is not allowed and we couldn't call it NewClient().  But since the convention is to name it New<struct-name>(), we could call it NewGroupMgmtClient().

For the option functions, I suggest:

- WithAddress(string) -- Don't specify "IP" address, since a host address is also accepted
- WithUser(string)
- WithTenantUser(string) -- Use this instead of WithUser() for a tenant user; it sets isTenant=true under the covers
- WithPassword(string)
- WithApiVersion(string)
- WithWaitForAsyncJobs()
- WithoutWaitForAsyncJobs()

I'm interested to hear what others think of this proposal.

rgcostea: This is a great idea.  One minor tweak I would offer is changing "WithAddress" to "WithHost".  I tend to see Host used more often than Address when dealing with SDKs.

raunakkumar: +1 i like the idea too

IrennaLumbuun: Should `NewNsGroupService()` be depreciated too? Right now the examples file calls on `NewNsGroupService()` and initialize it with ip, username, password, etc. If we were going to get rid of these parameters and replace them with optional parameters in the future, shouldn't  `NewNsGroupService()` also be replaced by another function that only accepts optional parameter (just like `NewGroupMgmtClient()`?

rkumpf: You are correct.  NewNsGroupService() is the interface that is used by client consumers.  I think we don't need to worry about compatibility for NewClient(); I had the two confused.  (The NewNsGroupService() changes are well hidden in the PR diff.)

So the question is, if we deprecate NewNsGroupService(), what do we name the replacement function?  I don't have a good answer for that.  Any suggestion?

Note that there is also a NewNsGroupService() function in the fakeservice package which should stay in sync.  I think that was missed in the changes so far.

IrennaLumbuun: I pushed a new commit addressing this change. For now, I use NewNimbleGroupService() as a placeholder name for the function.

p.s. changes on fakeService package is coming up :)

pkg/service/group_service_factory.go
```diff
@@ -16,69 +16,82 @@ type NsGroupService struct {
...
+	trustedOauthIssuerService         *TrustedOauthIssuerService
+	softwareVersionService            *SoftwareVersionService
+	ldapDomainService                 *LdapDomainService
+	networkConfigService              *NetworkConfigService
 }
 
 // NewNsGroupService - initializes NsGroupService
-func NewNsGroupService(ip, username, password, apiVersion string, synchronous bool) (gs *NsGroupService, err error) {
+// This function is depreciated. Call NewNimbleGroupService() to initialize a group service
+func NewNsGroupService(ip, username, password, apiVersion string, synchronous bool, clientOpts ...client.ClientOption) (gs *NsGroupService, err error) {
```
rkumpf: Remove clientOps.  No reason to enhance a deprecated function.  Let consumers use the new function to get new features.

pkg/service/group_service_factory.go
```diff
@@ -16,69 +16,82 @@ type NsGroupService struct {
...
 	}
 	return &NsGroupService{ip: ip, client: client}, nil
 }
 
+func NewNimbleGroupService(clientOpts ...client.ClientOption) (gs *NsGroupService, err error) {
+	groupMgmtClient, err := client.NewClient("", "", "", "", false, clientOpts...)
+	if err != nil {
+		return nil, err
+	}
+	if groupMgmtClient.ApiVersion != "v1" {
```
rkumpf: This check seems to belong in NewClient(), if that is where the client options get processed.

pkg/service/group_service_factory.go
```diff
@@ -16,69 +16,82 @@ type NsGroupService struct {
...
-	client, err := client.NewClient(ip, username, password, apiVersion, synchronous)
+	client, err := client.NewClient(ip, username, password, apiVersion, synchronous, clientOpts...)
 	if err != nil {
 		return nil, err
 	}
 	return &NsGroupService{ip: ip, client: client}, nil
 }
 
+func NewNimbleGroupService(clientOpts ...client.ClientOption) (gs *NsGroupService, err error) {
+	groupMgmtClient, err := client.NewClient("", "", "", "", false, clientOpts...)
```
rkumpf: We decided that NewClient() isn't really used by SDK consumers, right?  In that case we don't need to maintain compatibility.  We can either leave it with positional arguments or convert it to functional options.  I see no good reason to use a mixture of the two.  It might be simplest to keep all the functional options stuff here in the service and leave NewClient with positional arguments.  It isn't used externally anyway.

pkg/client/group_mgmt_client.go
```diff
@@ -60,7 +64,53 @@ type Argument struct {
 	JobId string `json:"job_id,omitempty"`
 }
 
-func newGroupMgmtClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) *GroupMgmtClient {
+type ClientOption func(*GroupMgmtClient)
```
rkumpf: Let's think about where these definitions belong.  If the NsGroupService is the consumer entry point, we'll want to locate these functions near it, probably in group_service_factory.go.  Maybe they are really ServiceOptions rather than ClientOptions.  (Honestly I find the client/service distinctions confusing.)

pkg/client/group_mgmt_client.go
```diff
@@ -60,7 +64,53 @@ type Argument struct {
...
+	}
+}
+
+func WithPassword(password string) func(*GroupMgmtClient) {
+	return func(groupMgmtClient *GroupMgmtClient) {
+		groupMgmtClient.Password = password
+	}
+}
+
+func WithApiVersion(version string) func(*GroupMgmtClient) {
```
rkumpf: Somewhere we need to define the default value as "v1" (which makes use of this function pretty unnecessary).

pkg/client/group_mgmt_client.go
```diff
@@ -60,7 +64,53 @@ type Argument struct {
...
+	return func(groupMgmtClient *GroupMgmtClient) {
+		groupMgmtClient.ApiVersion = version
+	}
+}
+
+func WithWaitForAsyncJobs() func(*GroupMgmtClient) {
```
rkumpf: What will we use as the default?  I suggest wait==true.

examples/fields.go
```diff
@@ -3,14 +3,14 @@ package main
 
 import (
 	"fmt"
-
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/client"
```
rkumpf: The need to add this import is a clue that we've defined the functions in the wrong place.  :)

examples/fields.go
```diff
@@ -3,14 +3,14 @@ package main
...
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
 )
 
 func main() {
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
+	groupService, err := service.NewNimbleGroupService(client.WithHost("1.1.1.1"), client.WithUser("xxx"), client.WithPassword("xxx"), client.WithApiVersion("v1"), client.WithoutWaitForAsyncJobs())
```
rkumpf: The waitForJobs setting was changed from true to false unintentionally?

examples/filters.go
```diff
@@ -12,7 +12,7 @@ import (
 func main() {
 
 	// login to Array, get groupService instance
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
+	groupService, err := service.NewNimbleGroupService(client.WithHost("1.1.1.1"), client.WithTenantUser("xxx"), client.WithPassword("xxx"), client.WithApiVersion("v1"), client.WithoutWaitForAsyncJobs())
```
rkumpf: The WithApiVersion() option should be unnecessary given the proper default. We can remove it from the examples.

examples/initiator_group.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func main() {
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
+	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false, false)
```
rkumpf: Remove the extra argument.

examples/restore.go
```diff
@@ -12,7 +12,7 @@ import (
 func main() {
 
 	// login to Array, get groupService instance
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
+	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false, false)
```
rkumpf: I guess all the examples should use non-deprecated functions.

examples/sdkservice.go
```diff
@@ -1,13 +1,13 @@
...
-	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
-	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/fakeservice"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/sdkprovider"
-	"fmt"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 )
 
 func getFakeService(ip, username, password, apiVersion string, synchronous bool) (sdkprovider.NsGroupService, error) {
```
rkumpf: Per my earlier comment, the fakeservice package should be kept in sync with the service changes.  Converting this example to the non-deprecated function should clarify it.

examples/sdkservice.go
```diff
@@ -47,21 +47,21 @@ func main() {
 		AgentType:      nimbleos.NsAgentTypeNone,
 	}
 
-	vol,_ := groupService.GetVolumeService().CreateVolume(newVolume)
+	vol, _ := groupService.GetVolumeService().CreateVolume(newVolume)
 	fmt.Printf("Fake volume %+v \n", vol)
 
 	// Get real service
-	groupService, _  = getRealService("10.21.4.79", "admin", "admin", "v1", true)
+	groupService, _ = getRealService("1.1.1.1", "xxx", "xxx", "v1", true, false)
```
rkumpf: This suggestion is not meant for this PR, but perhaps it can be taken up as a subsequent change if others think it worthwhile....

It seems inconvenient to have to edit the source to be able to run the examples.  Perhaps the host, user name, and password could be read from the environment instead.

The same issue exists in the test cases, for example volume_service_test.go.  Reading from the environment seems sensible for those, too.  Also the test cases should be changed to use the non-deprecated function.

pkg/client/group_mgmt_client.go
```diff
@@ -15,22 +15,26 @@ import (
...
 const (
-	groupURIFmt     = "https://%s:5392/%s"
-	clientTimeout   = time.Second * 60 // 1 Minute
-	maxLoginRetries = 1
-	jobTimeout      = time.Second * 300 // 5 Minute
-	jobPollInterval = 5 * time.Second   // Second
-	smAsyncJobId    = "SM_async_job_id"
+	groupURIFmtTenant = "https://%s:443/api/tenant/%s"
+	groupURIFmt       = "https://%s:5392/%s"
+	clientTimeout     = time.Second * 60 // 1 Minute
```
rkumpf: We could use time.Minute and remove the comment.  Similar for the other time constants here.

pkg/service/group_service_factory.go
IrennaLumbuun: hmm I'm not sure what you mean. Here's what I'm getting;
1. `NewNsGroupService() (depreciated)` calls on `NewClient()` so `NewClient` has to support positional arguments. 
2. The parameters of NewNimbleGroupService() is of type `ClientOption`, which returns a function that updates `GroupMgmtClient`. If NewClient() only accepts positional arguments, we have to find a way to extract the parameters from ClientOption, convert them to string or boolean, then pass them to `NewClient()`. That doesn't sound like a good solution.

I think what you're suggesting is to create another function in `service` that does what `NewClient()` does-- but receives optional arguments instead of positional arguments? Is that correct?

pkg/client/group_mgmt_client.go
IrennaLumbuun: hmm I'm thinking of defining the default value inside of `newGroupMgmtClient()` (see line 121-128). But if we have more than 1 version, wouldn't this function still be necessary? (for example, we set the default to be v1 but users can opt to use v2 instead)?

examples/fields.go
IrennaLumbuun: whoops, this is unintentional. I'm gonna fix it :D thanks!

pkg/client/group_mgmt_client.go
rkumpf: Yes, we should keep the function.  But we don't need to use it very often.  (v2 does not yet exist.)

pkg/client/v1/nimbleos/volume.go
IrennaLumbuun: Sorry I missed this. Working on a fix right now!

pkg/service/group_service_factory.go
rkumpf: Imagine for a moment that the only change we make in group_mgmt_client.go is to add another positional argument to NewClient().  All the functional options stuff can be handled here in group_service_factory.go.  I think I threw you off initially by putting my comment at the NewClient() code point when it should have been here.

With that premise, we could define a groupServiceOptions structure and a ServiceOption type that acts on it.  The NewNimbleGroupService() function can process the options into a groupServiceOptions object (lower case because it is not part of the external interface).  It can then call NewClient() with positional arguments that are extracted from the groupServiceOptions object.  With this approach, the default option values would be defined here in group_service_factory.go, which makes sense because they should be documented here at the interface an SDK consumer will use.

It might seem like we are creating duplication by defining a struct that is similar to GroupMgmtClient.  However, it is not really the same.  Notice how you needed to add Host and ApiVersion fields to GroupMgmtClient just to support the options processing.  I think the focus should be on defining an extensible interface that is clear and easy to use at the consumer entry point (here).  Subsequently perhaps we could find ways to streamline the interactions between the service package and the client package, but there's no rush there since the client interfaces are more easily changed.

examples/fields.go
```diff
@@ -3,14 +3,13 @@ package main
...
 	"fmt"
-
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
 )
 
 func main() {
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
+	groupService, err := service.NewNimbleGroupService(service.WithHost("1.1.1.1"), service.WithUser("xxx"), service.WithPassword("xxx"), service.WithWaitForAsyncJobs())
```
rkumpf: Let's keep the lines shorter.

examples/sdkservice.go
```diff
@@ -1,34 +1,34 @@
...
+	grpSvc, err := service.NewNimbleGroupService(clientOpts...)
 	return grpSvc, err
 }
 
 func main() {
 	arg := &param.GetParams{}
-	groupService, _ := getFakeService("1.1.1.1", "xxx", "xxx", "v1", true)
+	groupService, _ := getFakeService(service.WithHost("1.1.1.1"),
+		service.WithUser("xxx"), service.WithPassword("xxx"),
+		service.WithoutWaitForAsyncJobs())
```
rkumpf: Without -> With, or just remove this argument if it is the default.  Same thing at line 56.

pkg/client/group_mgmt_client.go
```diff
@@ -69,19 +71,31 @@ func newGroupMgmtClient(ipAddress, username, password, apiVersion string, waitOn
...
 }
 
 // NewClient instantiates a new client to communicate with the Nimble group
-func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs bool) (*GroupMgmtClient, error) {
+func NewClient(ipAddress, username, password, apiVersion string, waitOnJobs, isTenant bool) (*GroupMgmtClient, error) {
 	// Get resty client
-	groupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs)
+	groupMgmtClient := newGroupMgmtClient(ipAddress, username, password, apiVersion, waitOnJobs, isTenant)
+
+	if apiVersion != "v1" {
```
rkumpf, line 92: Put this check before the newGroupMgmtClient() call.  No point in doing that work if we'll throw it away.

pkg/fakeservice/group_service_factory.go
```diff
@@ -27,6 +28,16 @@ func NewNsGroupService(ip, username, password, apiVersion string, synchronous bo
 	return &NsGroupService{ip: ip}, nil
```
rkumpf, line 28: We should add a deprecation notice for this function.

pkg/service/group_service_factory.go
```diff
@@ -16,69 +15,142 @@ type NsGroupService struct {
...
+	fibreChannelInitiatorAliasService *FibreChannelInitiatorAliasService
+	groupService                      *GroupService
+	trustedOauthIssuerService         *TrustedOauthIssuerService
+	softwareVersionService            *SoftwareVersionService
+	ldapDomainService                 *LdapDomainService
+	networkConfigService              *NetworkConfigService
+}
+
+// groupServiceOption :
+type GroupServiceOption struct {
```
rkumpf: I'd say this should be plural, since it is an aggregate of all the options.

pkg/service/group_service_factory.go
```diff
@@ -16,69 +15,142 @@ type NsGroupService struct {
...
+}
+
+func WithoutWaitForAsyncJobs() func(*GroupServiceOption) {
+	return func(groupService *GroupServiceOption) {
+		groupService.WaitOnJob = false
+	}
 }
 
 // NewNsGroupService - initializes NsGroupService
+// This function is depreciated. Call NewNimbleGroupService() to initialize a group service
```
rkumpf: Typo *deprecated.

pkg/client/group_mgmt_client.go
```diff
@@ -69,19 +71,31 @@ func newGroupMgmtClient(ipAddress, username, password, apiVersion string, waitOn
...
+		return nil, fmt.Errorf("NewNimbleGroupService: unsupported %s sdk API version", apiVersion)
```
rkumpf: Remove "NewNimbleGroupService".  I suggest:

    return nil, fmt.Errorf("API version \"%s\" is not recognized", apiVersion)

rkumpf (commented):
This looks great, thank you.  Just a few minor items and it should be ready to go.

One of the commit messages was about updating the test cases, but it really updated the examples.  These test files could use attention:

- client/group_mgmt_client_test.go
- service/volume_service_test.go

examples/fields.go
rgcostea: Sorry.  I must have missed something.  Shouldn't this be:

`groupService, err := service.NewNimbleGroupService().WithHost("1.1.1.1").WithUser("xxx").WithPassword("xxx").WithWaitForAsyncJobs())`

IrennaLumbuun: @rgcostea I'm not the best person to answer this but I think it's a Golang thing (https://www.sohamkamani.com/golang/options-pattern/)

pkg/client/group_mgmt_client.go
IrennaLumbuun: Now that this function is moved to newGroupMgmtClient(), it can not return the error message (because newGroupMgmtClient only return 1 value that is *GroupMgmtClient). Should I just do  this instead:
```
fmt.Println("API version \"%s\" is not recognized", apiVersion)
return nil
```

examples/fields.go
rgcostea: I see.  It still seems cleaner to me to do the following:

`NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)`
versus
`NewNimbleGroupService(service.WithHost("1.1.1.1"), service.WithUser("xxx"), service.WithPassword("xxx"), service.WithWaitForAsyncJobs())`

It's more verbose than what it needs to be.  This is probably the point @rkumpf was making.

IrennaLumbuun: hmm but NewNsGroupService() is deprecated because it forces users to specify all 5 parameters. NewNimbleGroupService() allows them to specify only the parameters that they need (e.g. in this example, they don't have to specify the api version)

rgcostea: I see.  Sorry for the confusion.  This seems like it's the best we can do with golang.  Thanks for the pointers.

pkg/client/group_mgmt_client.go
rkumpf: Why is the check moved to newGroupMgmtClient()?  If necessary, we can add an error return value to that function.  Println is not a good option.

IrennaLumbuun: Oh, never mind. I misunderstood your other comment from last night. I'm gonna keep the check in this function

examples/initiator_group.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func main() {
-	groupService, err := service.NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", false)
+	groupService, err := service.NewNimbleGroupService(service.WithHost("1.1.1.1"), service.WithUser("xxx"), service.WithPassword("xxx"), service.WithoutWaitForAsyncJobs())
```
rkumpf: Add newlines like in fields.go in the other examples (and elsewhere?).

pkg/service/group_service_factory.go
```diff
@@ -67,64 +67,64 @@ type NsGroupService struct {
...
-	Password     string
-	isTenant     bool
-	ApiVersion   string
+// GroupServiceOptions :
+type GroupServiceOptions struct {
+	Host       string
+	WaitOnJob  bool
+	Username   string
+	Password   string
+	isTenant   bool
```
rkumpf: Why lower case here and upper case for the others?  (Sorry I didn't notice this before.)

pkg/client/group_mgmt_client.go
```diff
@@ -20,7 +20,7 @@ const (
 	clientTimeout     = time.Minute
 	maxLoginRetries   = 1
 	jobTimeout        = time.Minute * 5
```
rkumpf: I was going to let it go, but it is still bugging me...  Can we make it "5 * time.Minute" to be consistent with the line below?  Also remove the useless comment on the line below.

pkg/service/group_service_factory.go
IrennaLumbuun: I didn't realize this either. I'll make everything uppercase

pkg/client/group_mgmt_client.go
```diff
@@ -31,6 +32,7 @@ type GroupMgmtClient struct {
 	WaitOnJob    bool
 	Username     string
 	Password     string
+	isTenant     bool
```
rkumpf, line 35: Upper case here, too.

pkg/client/group_mgmt_client.go
rkumpf: Is another commit incoming for this one?

rkumpf (commented):
Don't forget to sign off each of the commits:
https://github.com/hpe-storage/nimble-golang-sdk/pull/53/checks?check_run_id=3127875340

Is there still an action outstanding for the _test.go files?  Any other open items?

IrennaLumbuun:
> Don't forget to sign off each of the commits:
> https://github.com/hpe-storage/nimble-golang-sdk/pull/53/checks?check_run_id=3127875340
> 
> Is there still an action outstanding for the _test.go files? Any other open items?

Yup, currently updating the test files

IrennaLumbuun:
I just noticed `volume_service_test.go` is spitting a bunch of errors. Let me look into it & I'll push a new commit once I figured it out

pkg/service/volume_service_test.go
```diff
@@ -27,7 +28,9 @@ type VolumeServiceTestSuite struct {
 
 func (suite *VolumeServiceTestSuite) config() *NsGroupService {
 
-	groupService, err := NewNsGroupService("1.1.1.1", "xxx", "xxx", "v1", true)
+	groupService, err := NewNimbleGroupService(WithHost("1.1.1.1"),
+	WithUser("xxx"), WithPassword("xxx"))
```
rkumpf: Check indentation

pkg/service/volume_service_test.go
```diff
@@ -71,41 +72,66 @@ func (suite *VolumeServiceTestSuite) SetupTest() {
...
+	// userObj := &nimbleos.User{
+	// 	TenantId: folder.TenantId,
+	// }
+	// user, _ = suite.nonTenantUserService.UpdateUser(tenantId, userObj)
+	// fmt.Println(*user.TenantId)
 
-	if err != nil {
-		suite.T().Errorf("NewGroupService(): Unable to connect to tenant group, err: %v", err.Error())
-	}
+	// // why is password NIL
```
raunakkumar: please remove the comments and commented out code.

raunakkumar (changes requested):

pkg/service/volume_service_test.go
```diff
@@ -37,23 +37,31 @@ type VolumeServiceTestSuite struct {
 	tenantFolderService			  		*FolderService
 }
 
-func (suite *VolumeServiceTestSuite) config() (*NsGroupService) {
+func (suite *VolumeServiceTestSuite) config() (*NsGroupService, *NsGroupService) {
```
rkumpf, line 45: Consider naming the return values and/or adding a function description.

pkg/service/volume_service_test.go
```diff
@@ -154,7 +126,10 @@ func (suite *VolumeServiceTestSuite) getDefaultVolumeOptions() *nimbleos.Volume
 	descriptionField := "This volume was created as part of a unit test"
 	var limitIopsField int64 = 256
 	var limitMbpsField int64 = 1
-	folder, _ := suite.nonTenantFolderService.GetFolderByName("default")
+
+	// TODO: replace <foldername> with your folder's name (owned by both admin and tenant)
```
rkumpf, line 135: Should we have a folder that is created by the test, with a predictable name?

pkg/service/volume_service_test.go
```diff
@@ -260,12 +235,19 @@ func (suite *VolumeServiceTestSuite) TestGetNonExistentVolumeByID() {
...
+
 	volume, err = suite.tenantVolumeService.GetVolumeById("06aaaaaaaaaaaaaaaa000000000000000000000000")
 	if err != nil {
 		suite.T().Errorf("TestGetNonExistentVolumeByID(): Unable to ge non-existent volume, err: %v", err.Error())
 		return
 	}
-	assert.Nil(suite.T(), volume)
+	/** Ideally, volume should be NIL (i.e. API returns HTTP 404 NOT FOUND)
+	 * But for some reason, it's returning HTTP 200, body = {}
+	 * So right not we test it by asserting volume.ID = nil.
```
rkumpf: Typo *now

pkg/service/volume_service_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Hewlett Packard Enterprise Development LP
+// Copyright 2020 - 2021 Hewlett Packard Enterprise Development LP
```
rkumpf: Remove the spaces around the hyphen, for consistency with the other files.

pkg/service/volume_service_test.go
```diff
@@ -87,53 +140,116 @@ func (suite *VolumeServiceTestSuite) getDefaultVolumeOptions() *nimbleos.Volume
...
+		FolderId:		folder.ID,
 	}
 	return newVolume
 }
 
-func (suite *VolumeServiceTestSuite) createVolume(volumeName string) *nimbleos.Volume {
-	volume, _ := suite.volumeService.GetVolumeByName(volumeName)
+func (suite *VolumeServiceTestSuite) nonTenantCreateVolume(volumeName string) *nimbleos.Volume {
+	// Create a volume for non tenant
+	volume, _ := suite.nonTenantVolumeService.GetVolumeByName(volumeName)
```
rkumpf: If we reuse an existing volume, it might not have been created by a tenant.  Maybe we aren't testing what we mean to test?  Similar thing in some of the other functions.

rkumpf (commented):
Overall the PR looks good to me, though I am not well-positioned to evaluate the volume_service_test.go changes.

pkg/service/volume_service_test.go
IrennaLumbuun: That's what I tried to do initially. The problem I ran into were:
1. Tenant doesn't have access to create a folder
2. Admin can create a folder, but this folder is not accessible to tenant
3. Admin can assign the folder to tenant, but it involves using an api with endpoint /api/cns/v1/users instead of the regular /api/tenant/v1/users and I got confused on how to incorporate that to the code.
So me and @raunakkumar  decided to leave it like this and assume tenant and folder id is given (especially since one of the optimization is to replace all placeholder values with environment variables anyway).

pkg/service/volume_service_test.go
IrennaLumbuun: mm I'm not sure what you mean.. There's 2 different createVolume functions: nonTenantCreateVolume() and tenantCreateVolume() so we can be sure which user does what

pkg/service/volume_service_test.go
rkumpf: I'll defer to Raunak.  It does seem like a concern that the SDK doesn't provide access to this API, though.

raunakkumar: We expect the Tenanadmin SDK to be used by the tenants. The array CLI already exposes a way to create tenants and folders. For now we haven't decided on exposing the admin APIs of the tenant proxy. We can always add that capability once we discuss it internally

pkg/service/volume_service_test.go
rkumpf: If a volume with the requested name exists before the "create" function is called, a new volume is not created.  So, the test might not verify that a tenant can create a volume.

pkg/service/volume_service_test.go
rkumpf: Gotcha.  Thanks for the explanation.

pkg/service/volume_service_test.go
IrennaLumbuun: ah okk. Just to double check -- you're referring to volume creation in SetupTest(), right?

pkg/service/volume_service_test.go
```diff
@@ -44,16 +44,16 @@ type VolumeServiceTestSuite struct {
  */
 func (suite *VolumeServiceTestSuite) config() (*NsGroupService, *NsGroupService) {
 
-	nonTenantGroupService, err := NewNimbleGroupService(WithHost("1.1.1.1"),
-		WithUser("xxx"), WithPassword("xxx"))
+	nonTenantGroupService, err := NewNimbleGroupService(WithHost("10.157.82.90"),
```
raunakkumar: remove ip address :)

IrennaLumbuun: whoops

pkg/service/volume_service_test.go
rkumpf: The volume creation is invoked in SetupTest(), yes, but that's not where I'm focusing.  Presumably the reason there are separate functions for tenant and non-tenant creation of volumes is to verify the behavior in both cases.  And it's not clear that is what is happening.  Maybe the GetVolumeByName() calls aren't needed in those functions, or maybe a pre-existing volume should cause a test failure.

IrennaLumbuun: got it!

rkumpf (approved):
Overall the PR looks good to me, though I am not well-positioned to evaluate the volume_service_test.go changes.

raunakkumar (approved):
lgtm as well
As @rkumpf pointed out, lets work on enhancing the volume_service_test.go and other places where we use parameters either from environment or through config file in a different PR

rkumpf (approved):

PR #54, opened 2021-07-29
Golang SDK Fields filtering is limited to fields of string type

IrennaLumbuun:


pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -4,33 +4,51 @@ package nimbleos
 
 // AccessControlRecord - Manage access control records for volumes.
```
rkumpf: I think this comment belongs just above the AccessControlRecord type declaration.

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -4,33 +4,51 @@ package nimbleos
...
 // Export AccessControlRecordFields for advance operations like search filter etc.
```
rkumpf: How about this?

// AccessControlRecordFields provides field names to use in filter parameters, for example.

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -4,33 +4,51 @@ package nimbleos
...
-var AccessControlRecordFields *AccessControlRecord
+var AccessControlRecordFields *AccessControlRecordStringFields
 
 func init() {
 	IDfield := "id"
+	ApplyTofield := "apply_to"
```
rkumpf: Shouldn't "field" be capitalized?

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -4,33 +4,51 @@ package nimbleos
...
 	ChapUserIdfield := "chap_user_id"
 	ChapUserNamefield := "chap_user_name"
 	InitiatorGroupIdfield := "initiator_group_id"
 	InitiatorGroupNamefield := "initiator_group_name"
+	Lunfield := "lun"
```
rkumpf: Is there a reason for these variables to be capitalized?  If it is easy to make them lower case, maybe we should.  I guess it isn't important.  A simple thing would be to put "field" at the beginning, like "fieldLun".

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -76,3 +94,27 @@ type AccessControlRecord struct {
 	// AccessProtocol - Access protocol of the volume.
 	AccessProtocol *NsAccessProtocol `json:"access_protocol,omitempty"`
 }
+
+// Struct for AccessControlRecordFields
+type AccessControlRecordStringFields struct {
```
rkumpf: I don't love this name, but I failed to come up with something better for this peculiar concept.  Maybe AccessControlRecordFieldHandles?

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -76,3 +94,27 @@ type AccessControlRecord struct {
 	// AccessProtocol - Access protocol of the volume.
 	AccessProtocol *NsAccessProtocol `json:"access_protocol,omitempty"`
 }
+
+// Struct for AccessControlRecordFields
```
rkumpf: I suggest:

// AccessControlRecordStringFields provides a string representation for each AccessControlRecord field.

examples/volume.go
```diff
@@ -50,6 +50,21 @@ func main() {
...
+		*nimbleos.VolumeFields.Size,
+		*nimbleos.VolumeFields.LimitMbps,
+	}
+	requestParams.WithFields(fieldList)
+	_, err  = volSvc.GetVolumes(requestParams)
+	if err != nil{
+		fmt.Println("Error: get volume with params. Message: %s", err)
+	}
+
 	fmt.Println(volume)
```
rkumpf: Move this up next to the CreateVolume call?

examples/volume.go
```diff
@@ -50,6 +50,21 @@ func main() {
...
+	// get volume with params
+	requestParams := new(param.GetParams)
+	fieldList := []string{
+		*nimbleos.VolumeFields.ID,
+		*nimbleos.VolumeFields.Name,
+		*nimbleos.VolumeFields.Size,
+		*nimbleos.VolumeFields.LimitMbps,
+	}
+	requestParams.WithFields(fieldList)
+	_, err  = volSvc.GetVolumes(requestParams)
```
rkumpf: Should we print the result or something?  Maybe just the first volume in the list.

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -4,33 +4,51 @@ package nimbleos
...
+	PeIdsfield := "pe_ids"
+	Snaplunsfield := "snapluns"
+	CreationTimefield := "creation_time"
+	LastModifiedfield := "last_modified"
+	AccessProtocolfield := "access_protocol"
 
-	AccessControlRecordFields = &AccessControlRecord{
+	AccessControlRecordFields = &AccessControlRecordStringFields{
 		ID:                 &IDfield,
+		ApplyTo:            &ApplyTofield,
```
rkumpf: Now that we've broken the relationship with the original AccessControlRecord struct, there doesn't seem to be a reason to use pointers anymore.  If we change all the fields from *string to string, we wouldn't need the temporary variables.  We can just use string literals inline here.

pkg/fakeservice/group_service_factory.go
```diff
@@ -29,7 +29,7 @@ func NewNsGroupService(ip, username, password, apiVersion string, synchronous bo
 }
 
 func NewNimbleGroupService(serviceOpts ...service.ServiceOptions) (gs *NsGroupService, err error) {
-	fakeGroupServiceOption := &service.GroupServiceOption{}
+	fakeGroupServiceOption := &service.GroupServiceOptions{}
```
rkumpf: Maybe make this variable name plural, to match.

pkg/fakeservice/group_service_factory.go
```diff
@@ -29,7 +29,7 @@ func NewNsGroupService(ip, username, password, apiVersion string, synchronous bo
 }
 
 func NewNimbleGroupService(serviceOpts ...service.ServiceOptions) (gs *NsGroupService, err error) {
```
rkumpf: Seeing it in this context, it looks like service.ServiceOptions should be renamed to service.ServiceOption.

IrennaLumbuun:
Sorry, had to force push a new commit because I had some trouble rebasing / adding signatures to the previous commits.

pkg/client/v1/nimbleos/access_control_record.go
IrennaLumbuun: Yeah, I'll just add a newline to separate it from the next comment

pkg/client/v1/nimbleos/access_control_record.go
IrennaLumbuun: hmm yeah that looks right. I'll capitalize it

pkg/client/v1/nimbleos/access_control_record.go
IrennaLumbuun: Not sure why it needs to be capitalized, I was just following the code in the master branch. Sure, we can change it to `field{UpperCamelCaseOfTheFieldName}`

pkg/client/v1/nimbleos/access_control_record.go
IrennaLumbuun: As per your other comment, I'm changing it to `field{VariableName}`

pkg/client/v1/nimbleos/access_control_record.go
IrennaLumbuun: Yeah, it took me a while to come up with a name as well. 
Tagging the others to join the conversation -- @raunakkumar @rgcostea @suneeth51 @vidhutsingh

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: Sorry if I wasn't clear.  I'm saying this comment should be attached to the AccessControlRecord struct definition.  It is less visible if it is just free-floating somewhere in this file.

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: We won't need the field variables at all if we remove the pointers from the struct.

pkg/client/v1/nimbleos/access_control_record.go
raunakkumar: i don't have a better suggestion too. `AccessControlRecordFieldHandles`  works for me

pkg/client/v1/nimbleos/access_control_record.go
raunakkumar: Yes, that would be ideal. Would that cause a lot of code churn?

pkg/client/v1/nimbleos/access_control_record.go
rkumpf: Note to other reviewers: Making the values pure strings without pointers will break drop-in compatibility.  Upgrading to the new SDK version will require removing '*' characters where the fields are used (such as in examples/volume.go above).  I believe this is an appropriate trade-off because:

1. The field list used in the request parameters is of type []string.  These fields are defined mainly for that use.  We shouldn't require pointer dereferences to use them.
2. Existing use of the fields specifier was largely degraded prior to this fix, since it could only be used for string fields.  Therefore the impact might be minimal.

IrennaLumbuun: Ah that's nice!

raunakkumar: I think the change is worth it. We can add in the release note that this a breaking feature for existing SDK users and how to overcome it. Will let others chime in as well

pkg/client/v1/nimbleos/access_control_record.go
IrennaLumbuun: ah ok got it.

pkg/fakeservice/group_service_factory.go
IrennaLumbuun: I think we had this conversation in the previous pull request. We changed it from singular to plural because it contains all the services options, i.e. this chunk of code
 
```
// GroupServiceOptions :
type GroupServiceOptions struct {
	Host       string
	WaitOnJob  bool
	Username   string
	Password   string
	IsTenant   bool
	ApiVersion string
}
```

rkumpf: You are right about the GroupServiceOptions name.  We discussed it here: https://github.com/hpe-storage/nimble-golang-sdk/pull/53#discussion_r673724104.

This comment is about ServiceOptions, which is a type declaration for a function that handles a single option.  I'm suggesting changing this:

`type ServiceOptions func(*GroupServiceOptions)`

to this:

`type ServiceOption func(*GroupServiceOptions)`

It seems strange for the NewNimbleGroupService() function to take a list of option parameters, each of which is plural.

IrennaLumbuun: ah okay I see! Thanks for the clarification!

vidhutsingh (changes requested):
This change will break the SDK. we cannot make this change. Pointers are intentionally used here so that at the time of json encoding/decoding attributes values are passed properly to storage rest server.

examples/fields.go
```diff
@@ -10,10 +10,9 @@ import (
 
 func main() {
 	groupService, err := service.NewNimbleGroupService(
-		service.WithHost("1.1.1.1"),
-		service.WithUser("xxx"),
-		service.WithPassword("xxx"),
-		service.WithWaitForAsyncJobs())
+		service.WithHost("10.157.82.90"),
```
vidhutsingh: Please  remove the credentials and IP address as it is public repo

examples/fields.go
```diff
@@ -51,16 +50,16 @@ func main() {
...
+		nimbleos.VolumeFields.Name,
+		nimbleos.VolumeFields.PerfpolicyName,
 	}
 	// apply attributes
 	filter.WithFields(volAttrList)
 
 	// create a filter
 	sf := &param.SearchFilter{
-		FieldName: nimbleos.VolumeFields.Name,
+		FieldName: param.NewString(nimbleos.VolumeFields.Name),
```
rkumpf: Hmmm...  I didn't see this coming.  Maybe we could use &nimbleos.VolumeFields.Name without the param.NewString() function.  But we might be better off using the SearchFilter Init() function.

I see the same thing shows up in a bunch of the services.

I'm tempted to dig into why FieldName is a pointer, but I'll resist....

examples/volume.go
```diff
@@ -51,6 +50,22 @@ func main() {
 		return
 	}
 	fmt.Println(volume)
+	// get volume with params
```
rkumpf, line 63: Add a blank line above.

pkg/service/active_directory_membership_service.go
```diff
@@ -74,7 +74,7 @@ func (svc *ActiveDirectoryMembershipService) GetActiveDirectoryMembershipById(id
 func (svc *ActiveDirectoryMembershipService) GetActiveDirectoryMembershipByName(name string) (*nimbleos.ActiveDirectoryMembership, error) {
 	params := &param.GetParams{
 		Filter: &param.SearchFilter{
-			FieldName: nimbleos.VolumeFields.Name,
+			FieldName: param.NewString(nimbleos.VolumeFields.Name),
```
rkumpf: It doesn't seem right that we are using VolumeFields in all these places that are unrelated to volumes.  It should be &nimbleos.ActiveDirectoryMembershipFields.Name, right?

rkumpf:
> This change will break the SDK. we cannot make this change. Pointers are intentionally used here so that at the time of json encoding/decoding attributes values are passed properly to storage rest server.

Can you point out specifically where you see a problem?  The non-pointer strings are used in the definition of a new struct that is not directly used for serialization/deserialization.

vidhutsingh:
> > This change will break the SDK. we cannot make this change. Pointers are intentionally used here so that at the time of json encoding/decoding attributes values are passed properly to the storage rest server.
> 
> Can you point out specifically where you see a problem? The non-pointer strings are used in the definition of a new struct that is not directly used for serialization/deserialization.

Sorry, I missed seeing that there is a wrapper implementation around nimble objects e.g AccessControlRecordFieldHandles 
As long as we are not serializing/deserializing we should be good. 

@irenna-lumbuun , please scan the code and make the filter changes where ever search filters are used. It should be in the service folder mostly.

pkg/service/active_directory_membership_service.go
IrennaLumbuun: Hmm I'm not sure; it was already there when I branched from SM's default branch. (I pasted a snippet below from `ServicesTemplate.ftl`). Are you sure we need to change it?
```
// Get${DataObjectName}ByName - method returns a pointer "${DataObjectName}"
func (svc *${DataObjectName}Service) Get${DataObjectName}ByName(name string) (*nimbleos.${DataObjectName}, error) {
    params := &param.GetParams{
        Filter: &param.SearchFilter{
            FieldName: param.NewString(nimbleos.VolumeFields.Name),
            Operator:  param.EQUALS.String(),
            Value:     name,
        },
    }
    ${DataObjectName?uncap_first}Resp, err := svc.objectSet.GetObjectListFromParams(params)
    if err != nil {
        return nil, err
    }

    if len(${DataObjectName?uncap_first}Resp) == 0 {
        return nil, nil
    }

    return ${DataObjectName?uncap_first}Resp[0],nil
    </#if>
}
```

examples/fields.go
IrennaLumbuun: I'm not sure what you mean.. Should these be
```
sf := &param.SearchFilter{
	    FieldName: &nimbleos.VolumeFields.Name,
            ...
        }
```
?

pkg/service/active_directory_membership_service.go
rkumpf: I understand this issue was not introduced by your PR.  Am I sure we need to change it?  No, it presumably works fine the way it is.  But it seems wrong philosophically to filter a GET request on a field from a different object type.  I think we should change VolumeFields.Name to ${DataObjectName}Fields.Name.  If that change is found to cause a problem somewhere, it will be interesting to understand why.

examples/fields.go
rkumpf: Right, that should work.  I noticed there is an Init() function for SearchFilter which takes a string rather than a string pointer.  I'm not sure exactly how it is intended to be used.  It might not be worth considering here.

rkumpf: Use & instead of param.NewString() in the examples, too.

examples/filters.go
```diff
@@ -13,8 +13,8 @@ func main() {
 	// login to Array, get groupService instance
 	groupService, err := service.NewNimbleGroupService(
 		service.WithHost("1.1.1.1"),
-		service.WithTenantUser("xxx"),
-		service.WithPassword("xxx"))
+ 		service.WithUser("xxx"),
```
rkumpf: Was an extra space character added at the beginning of the line?  Also on the next line and in some of the other example files.

examples/sdkservice.go
```diff
@@ -50,8 +52,10 @@ func main() {
 	fmt.Printf("Fake volume %+v \n", vol)
 
 	// Get real service
-	groupService, _ = getRealService(service.WithHost("1.1.1.1"),
-		service.WithUser("xxx"), service.WithPassword("xxx"))
+	groupService, _ = getRealService(
+		service.WithHost("10.157.82.90"),
```
rkumpf: D'oh!

examples/volume.go
```diff
@@ -12,9 +12,8 @@ import (
 func main() {
 	groupService, err := service.NewNimbleGroupService(
 		service.WithHost("1.1.1.1"),
-		service.WithUser("xxx"),
-		service.WithPassword("xxx"),
-		service.WithoutWaitForAsyncJobs())
```
rkumpf: Is the change to asynchronous jobs intentional?

Is there an example anywhere that shows how to use the SDK to process an async job?  I.e., to interpret the response and recognize that the job is incomplete and then poll for its completion.  I would love to see that.

For now this change to use synchronous wait is probably a good idea.  I suggest adding a separate item to the backlog to provide an async job example, if reasonably feasible.

examples/volume.go
```diff
@@ -48,9 +47,32 @@ func main() {
 
 	if err != nil {
 		fmt.Printf("Failed to create volume, err ", err)
-		return
```
rkumpf, line 51: Why is the return statement removed?  If volume creation fails, the subsequent actions will presumably also fail.

pkg/client/v1/nimbleos/access_control_record.go
```diff
@@ -2,38 +2,35 @@
 
 package nimbleos
 
-// AccessControlRecord - Manage access control records for volumes.
-// Export AccessControlRecordFields for advance operations like search filter etc.
-var AccessControlRecordFields *AccessControlRecord
+// Export AccessControlRecordFields provides field names to use in filter parameters, for example.
```
rkumpf: I don't think we want the word "Export" here.

pkg/service/group_service_factory.go
```diff
@@ -159,7 +159,7 @@ func (gs *NsGroupService) LogoutService() error {
 	// apply filter on session token
 	sFilter := &param.GetParams{
 		Filter: &param.SearchFilter{
-			FieldName: nimbleos.TokenFields.SessionToken,
+			FieldName: param.NewString(nimbleos.TokenFields.SessionToken),
```
rkumpf: Use & instead of param.NewString().

examples/initiator_group.go
```diff
@@ -89,15 +89,24 @@ func main() {
 		}
 	}
 
+	// get igroup by name
+	igroup, err = igroupSvc.GetInitiatorGroupByName("sdkigroup")
+	if err != nil {
+		fmt.Printf("Failed to get igroup by name, err ", err)
```
rkumpf: The message formatting doesn't look right.  Same thing in restore.go.

examples/volume.go
```diff
@@ -51,6 +51,31 @@ func main() {
...
+ 		nimbleos.VolumeFields.ID,
+ 		nimbleos.VolumeFields.Name,
+ 		nimbleos.VolumeFields.Size,
+ 		nimbleos.VolumeFields.LimitMbps,
+ 	}
+ 	requestParams.WithFields(fieldList)
+ 	volumeList, err := volSvc.GetVolumes(requestParams)
+
+	if err != nil{
+		fmt.Println("Error: get volume with params. Message: %s", err)
```
rkumpf: I don't think the '%s' will work in a Println call.

examples/volume.go
```diff
@@ -51,6 +51,31 @@ func main() {
...
+ 	fieldList := []string{
+ 		nimbleos.VolumeFields.ID,
+ 		nimbleos.VolumeFields.Name,
+ 		nimbleos.VolumeFields.Size,
+ 		nimbleos.VolumeFields.LimitMbps,
+ 	}
+ 	requestParams.WithFields(fieldList)
+ 	volumeList, err := volSvc.GetVolumes(requestParams)
+
+	if err != nil{
```
rkumpf: Add a space before the bracket.

examples/filters.go
rkumpf: Whitespace can be difficult to see, but the green square in the diff looks like an added space.

examples/volume.go
rkumpf: @raunakkumar, thoughts about adding a backlog item to create an async job example?

examples/filters.go
IrennaLumbuun: Oh isn't that because WithTenantUser vs WithUser?

examples/fields.go
```diff
@@ -45,7 +45,7 @@ func main() {
 	// create volume
 	volume, err := volSvc.CreateVolume(newVolume)
 	if err != nil {
-		fmt.Println("Failed to create volume, ", err)
+		fmt.Println("Failed to create volume, err: %v", err)
```
rkumpf: I think this one was correct before.  You might want to use Printf with this change.

Check out the Go Playground for an easy way to experiment with how the print functions work: https://play.golang.org/.

examples/initiator_group.go
```diff
@@ -92,7 +92,7 @@ func main() {
 	// get igroup by name
 	igroup, err = igroupSvc.GetInitiatorGroupByName("sdkigroup")
 	if err != nil {
-		fmt.Printf("Failed to get igroup by name, err ", err)
+		fmt.Printf("Failed to get igroup by name, err: %v", err)
```
rkumpf: Since Printf doesn't include a trailing newline, it is common to add \n to the end of the message.

examples/filters.go
rkumpf: I don't think so.  Notice how the password line also shows a change.  Similar in initiator_group.go and restore.go.

examples/fields.go
IrennaLumbuun: ah ok I didn't notice some of them were println and the others were printf

vidhutsingh (approved):

rkumpf (approved):

raunakkumar (approved):

PR #55, opened 2021-08-03
Change input values to environment variables

IrennaLumbuun:
Update test files and example files with environment variables

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,14 +6,6 @@ The SDK provides Go modules to interact with the HPE Nimble Storage REST API. Th
 # Synopsis
 
 Examples are available in [examples](examples). This is a brief Go program that creates a 5GiB volume on the specified array.
-In order to run [examples](examples) files, the following environment variables need to be set:
-	- SDK_TARGET_HOST				// Managment hostname or IP of array
-	- SDK_TARGET_USER				// User (non-tenant) username
-	- SDK_TARGET_USER_PASSWORD		// User (non-tenant) password
-	- SDK_TARGET_TENANT_USER		// Tenant username
-	- SDK_TARGET_TENANT_PASSWORD	// Tenant password
-	- SDK_TARGET_TENANT_FOLDER		// The id of tenant's folder
-
 
 ```
 package main
--- a/README.md
+++ b/README.md
@@ -6,6 +6,14 @@ The SDK provides Go modules to interact with the HPE Nimble Storage REST API. Th
 # Synopsis
 
 Examples are available in [examples](examples). This is a brief Go program that creates a 5GiB volume on the specified array.
+In order to run [examples](examples) files, the following environment variables need to be set:
+	- SDK_TARGET_HOST				// Managment hostname or IP of array
+	- SDK_TARGET_USER				// User (non-tenant) username
+	- SDK_TARGET_USER_PASSWORD		// User (non-tenant) password
+	- SDK_TARGET_TENANT_USER		// Tenant username
+	- SDK_TARGET_TENANT_PASSWORD	// Tenant password
+	- SDK_TARGET_TENANT_FOLDER		// The id of tenant's folder
+
 
 ```
 package main
@@ -20,12 +28,10 @@ import (
 func main() {
 
 	// Create new group service
-	groupService, err := service.NewNsGroupService(
- 		"192.168.1.1",              // Managment hostname or IP of array
- 		"admin",                    // Username
-		"admin",                    // Password
-		"v1",                       // REST API version
-		true)                       // Perform operations synchronous
+	groupService, err := service.NewNimbleGroupService(
+		service.WithHost("192.168.1.1"),	// Managment hostname or IP of array
+		service.WithUser("admin"),			// Username
+		service.WithPassword("admin"))		// Password
 
 	if err != nil {
 		fmt.Printf("Unable to connect to group, %+v\n", err.Error())
--- a/examples/fields.go
+++ b/examples/fields.go
@@ -16,11 +16,10 @@ func main() {
 
 	if host == "" || user == "" || password == "" {
 		fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_PASSWORD")
-		fmt.Println(
-		`Usage:
-		- SDK_TARGET_HOST				// Managment hostname or IP of array
-		- SDK_TARGET_USER				// User (non-tenant) username
-		- SDK_TARGET_USER_PASSWORD		// User (non-tenant) password`)
+		fmt.Println("Usage:")
+		fmt.Println("SDK_TARGET_HOST - Management hostname or IP of array")
+		fmt.Println("SDK_TARGET_USER - User (non-tenant) username")
+		fmt.Println("SDK_TARGET_USER_PASSWORD - User (non-tenant) password")
 		os.Exit(1)
 	}
 
--- a/examples/fields.go
+++ b/examples/fields.go
@@ -3,130 +3,32 @@ package main
 
 import (
 	"fmt"
-	"flag"
 	"os"
-	"strings"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
 )
 
-func handleOptionError(target string){
-	fmt.Printf("ERROR: Environment variable SDK_TARGET_%s is not defined.\n", strings.ToUpper(target))
-	os.Exit(1)
-}
-
-/*
-	This function returns: host, isTenant (indicates whether we're using regular user or tenantUser), username, and password
-*/
-func parseOptionalArgument()(string, bool, string, string){
-	// define flags
-	host := flag.String("host", "", "SDK Target Host")
-	user := flag.String("user", "", "SDK Target User")
-	password := flag.String("user-password", "", "SDK Target User Password")
-	tenant := flag.String("tenant", "", "SDK Target Tenant")
-	tenantPassword := flag.String("tenant-password", "", "SDK Target Tenant Password")
-
-	flag.Parse()
-	fmt.Println(*host, *user, *password, *tenant, *tenantPassword);
-
-	if *host == "" {
-		*host = os.Getenv("SDK_TARGET_HOST")
-	}
-
-	if *user == "" {
-		*user = os.Getenv("SDK_TARGET_USER")
-	}
-
-	if *password == "" {
-		*password = os.Getenv("SDK_TARGET_USER_PASSWORD")
-	}
-
-	if *tenant == "" {
-		*tenant = os.Getenv("SDK_TARGET_TENANT_USER")
-	}
-
-	if *tenantPassword == "" {
-		*tenantPassword = os.Getenv("SDK_TARGET_TENANT_PASSWORD")
-	}
-
-	// Host has to be defined
-	if *host == "" {
-		handleOptionError("host")
-	}
-
-	// Either (user & password) or (tenant & tenantPassword) has to be defined
-	isTenant := false
-
-	if *user == "" && *password== "" && *tenant == "" && *tenantPassword == "" {
-		fmt.Println("ERROR: Must define either a user or a tenant.")
-		os.Exit(1)
-	}
-
-	// Initalize with user instead of tenant
-	if *tenant == "" && *tenantPassword == "" {
-		// Error handling
-		if (*user != "" || *password != "") {
-			if *user != "" && *password == "" {
-				handleOptionError("user_password")
-			}
-			if *user == "" && *password != "" {
-				handleOptionError("user")
-			}
-		}
-		return *host, isTenant, *user, *password
-	}
-
-	// Initialize with a tenant
-	if *user == "" && *password == "" {
-		isTenant = true
-		// Error handling
-		if (*tenant != "" || *tenantPassword != ""){
-			if *tenant != "" && *tenantPassword == "" {
-				handleOptionError("tenant_password")
-			}
-			if *tenant == "" && *tenantPassword != "" {
-				handleOptionError("tenant_user")
-			}
-		}
-		return *host, isTenant, *tenant, *tenantPassword
-	}
-
-	// if all user, password, tenant, and tenantPassword are initialized
-	// default to regular user
-	fmt.Println("Both tenant's password & username and user's password & username are defined")
-	fmt.Println("User's username and password are used for service initialization")
-	return *host, isTenant, *user, *password
-}
-
-func initializeGroupService(host string, isTenant bool, username, password string) (*service.NsGroupService) {
-	if isTenant {
-		groupService, err := service.NewNimbleGroupService(
-			service.WithHost(host),
-			service.WithTenantUser(username),
-			service.WithPassword(password))
-		if err != nil {
-			fmt.Printf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
+func checkEnvironmentVariableExists(){
+	if (os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "") {
+			fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_USER_PASSWORD");
 			os.Exit(1)
 		}
-		return groupService
-	} else {
-		groupService, err := service.NewNimbleGroupService(
-			service.WithHost(host),
-			service.WithUser(username),
-			service.WithPassword(password))
-		if err != nil {
-			fmt.Printf("NewGroupService(): Unable to connect to group, err: %v", err.Error())
-			os.Exit(1)
-		}
-		return groupService
-	}
 }
 
 func main() {
-	host, isTenant, username, password := parseOptionalArgument()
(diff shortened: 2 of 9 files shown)


raunakkumar (approved):
lgtm. Maybe add an example of these environment variable in Test Readme?

examples/fields.go
```diff
@@ -3,16 +3,17 @@ package main
...
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
 )
 
 func main() {
 	groupService, err := service.NewNimbleGroupService(
-		service.WithHost("1.1.1.1"),
-		service.WithUser("xxx"),
-		service.WithPassword("xxx"))
+		service.WithHost(os.Getenv("GO_SDK_TEST_HOST")),
```
rkumpf: For these examples with a main() function, is there a convenient way to invoke them with options?  Perhaps it would be appropriate to define flags to allow the values to be specified on invocation, and fall back on the environment variables if no options are given.

pkg/client/group_mgmt_client_test.go
```diff
@@ -21,7 +22,7 @@ func TestNewClient(t *testing.T) {
 
 	// Create client
 	var err error
-	client, err = NewClient("x.x.x.x", "xxx", "xxx", "v1", true, false)
+	client, err = NewClient(os.Getenv("GO_SDK_TEST_HOST"), os.Getenv("GO_SDK_TEST_USER"), os.Getenv("GO_SDK_TEST_USER_PASSWORD"), "v1", true, false)
```
rkumpf: The test should fail gracefully if the environment variables are not defined.  We could produce a message detailing the environmental requirements.

examples/fields.go
```diff
@@ -3,16 +3,17 @@ package main
...
+		service.WithUser(os.Getenv("GO_SDK_TEST_USER")),
```
rkumpf: I wonder if "GO" is important to include in the env var names, or even "SDK".  It is tempting to use "NIMBLE", like "NIMBLE_TEST_USER", but we might be better off avoiding the branding even though that term is all over the SDK.

Also, these are examples rather than tests, so maybe TEST isn't right either.

How about SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_PASSWORD?

rkumpf (commented):
Oh dear, I did not mean to cause a code explosion.  It looks like the flags are adding much more complexity than they are worth.  Sorry about that.  Can we get back to something relatively simple?  For these little examples, it should be good enough to just notice if any of the environment variables is not defined and print a message listing all the required settings.  Same thing in the test files.  And in the examples do we need to add tenant options?  If so, maybe a single flag is enough instead of separate options for tenant and non-tenant users.

IrennaLumbuun:
> Oh dear, I did not mean to cause a code explosion. It looks like the flags are adding much more complexity than they are worth. Sorry about that. Can we get back to something relatively simple? For these little examples, it should be good enough to just notice if any of the environment variables is not defined and print a message listing all the required settings. Same thing in the test files. And in the examples do we need to add tenant options? If so, maybe a single flag is enough instead of separate options for tenant and non-tenant users.

Just making sure.. so, are we getting rid of all flags (-host, -user, etc) and replace it with a single flag that tells us whether it's tenant or regular user (i.e. -isTenant)?

rkumpf:
> Just making sure.. so, are we getting rid of all flags (-host, -user, etc) and replace it with a single flag that tells us whether it's tenant or regular user (i.e. -isTenant)?

If we need the ability to connect as a tenant user in the example, I would make it an env var rather than a flag.  I'm not sure why the examples need a tenant option, though.

The use of flags seems to have added hundreds of lines of code which kind of overwhelm the examples, so I would eliminate all the flags unless they can be added succinctly.

examples/fields.go
```diff
@@ -6,16 +6,28 @@ import (
...
+	if os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "" {
+		fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_USER_PASSWORD")
+		os.Exit(1)
+	}
+}
+
 func main() {
+	checkEnvironmentVariableExists()
```
rkumpf: I suggest reading the env vars just once.  For example:

	host := os.Getenv("SDK_TARGET_HOST")
	user := os.Getenv("SDK_TARGET_USER")
	password := os.Getenv("SDK_TARGET_PASSWORD")

	if host == "" || user == "" || password == "" {
		fmt.Println("ERROR: These environment variables must be defined: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_PASSWORD.")
		os.Exit(1)
	}

Then use the same variables below.

examples/fields.go
```diff
@@ -6,16 +6,28 @@ import (
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/client/v1/nimbleos"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/param"
 	"github.com/hpe-storage/nimble-golang-sdk/pkg/service"
+	"os"
 )
 
+func checkEnvironmentVariableExists() {
+	if os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "" {
```
rkumpf: SDK_TARGET_PASSWORD seems clear enough.

examples/filters.go
```diff
@@ -40,6 +50,6 @@ func main() {
 	sfilter.WithSearchFilter(f)
 
 	//GetVolumes
-	volumes, err := volSvc.GetVolumes(sfilter)
+	volumes, _ := volSvc.GetVolumes(sfilter)
```
rkumpf: It would be better to check and print the error.

pkg/service/volume_service_test.go
```diff
@@ -37,23 +37,35 @@ type VolumeServiceTestSuite struct {
...
 }
 
+func checkEnvironmentVariableExists() {
+	if os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "" ||
+		os.Getenv("SDK_TARGET_TENANT_USER") == "" ||
+		os.Getenv("SDK_TARGET_TENANT_PASSWORD") == "" ||
+		os.Getenv("SDK_TARGET_TENANT_FOLDER") == "" {
+		suite.T().Errorf("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_USER_PASSWORD, SDK_TARGET_TENANT_USER, SDK_TARGET_TENANT_PASSWORD, SDK_TARGET_TENANT_FOLDER")
```
rkumpf: Use Error() instead of Errorf()?

pkg/service/volume_service_test.go
```diff
@@ -37,23 +37,35 @@ type VolumeServiceTestSuite struct {
 	tenantFolderService             *FolderService
 }
 
+func checkEnvironmentVariableExists() {
+	if os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "" ||
+		os.Getenv("SDK_TARGET_TENANT_USER") == "" ||
```
rkumpf: Is it obvious what all these env vars mean and how to set them?  If not, the error message could be expanded and/or documentation added somewhere.

IrennaLumbuun: hmm should the documentation be added in Readme instead? The error message could be
```
ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_PASSWORD
See 'Readme' for usage
```

examples/fields.go
```diff
@@ -6,16 +6,31 @@ import (
...
+		os.Exit(1)
+	}
+}
+
 func main() {
+	host := os.Getenv("SDK_TARGET_HOST")
+	user := os.Getenv("SDK_TARGET_USER")
+	password := os.Getenv("SDK_TARGET_PASSWORD")
+
+	checkEnvironmentVariableExists(host, user, password)
```
rkumpf: Is there a reason to put this in a separate function?  I thought it could stay inline like I showed in my previous comment.

pkg/service/volume_service_test.go
rkumpf: The README file is referring to the examples directory.  The test requirements seem to be different.  They are two different audiences.

pkg/client/group_mgmt_client_test.go
```diff
@@ -17,11 +18,20 @@ var (
...
+	if (os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "") {
+			fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_USER_PASSWORD");
+			os.Exit(1)
+		}
+}
 
+func TestNewClient(t *testing.T) {
+	checkEnvironmentVariableExists();
```
rkumpf: Add this check to TestListGetOrPost() also?  There are other ways to consolidate, but probably not worth the effort.

README.md
```diff
@@ -6,6 +6,14 @@ The SDK provides Go modules to interact with the HPE Nimble Storage REST API. Th
 # Synopsis
 
 Examples are available in [examples](examples). This is a brief Go program that creates a 5GiB volume on the specified array.
+In order to run [examples](examples) files, the following environment variables need to be set:
+	- SDK_TARGET_HOST				// Managment hostname or IP of array
+	- SDK_TARGET_USER				// User (non-tenant) username
+	- SDK_TARGET_USER_PASSWORD		// User (non-tenant) password
+	- SDK_TARGET_TENANT_USER		// Tenant username
```
rkumpf: This list should include only the settings that are required for the examples.

examples/fields.go
```diff
@@ -6,16 +6,31 @@ import (
...
+	user := os.Getenv("SDK_TARGET_USER")
+	password := os.Getenv("SDK_TARGET_PASSWORD")
+
+	if host == "" || user == "" || password == "" {
+		fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_PASSWORD")
+		fmt.Println(
+		`Usage:
+		- SDK_TARGET_HOST				// Managment hostname or IP of array
+		- SDK_TARGET_USER				// User (non-tenant) username
+		- SDK_TARGET_USER_PASSWORD		// User (non-tenant) password`)
```
rkumpf: SDK_TARGET_PASSWORD

It's probably better to avoid using tabs in the message.  Separate Println statements might help readability, without having to worry about conforming indentation.

examples/fields.go
```diff
@@ -6,16 +6,31 @@ import (
...
 func main() {
+	host := os.Getenv("SDK_TARGET_HOST")
+	user := os.Getenv("SDK_TARGET_USER")
+	password := os.Getenv("SDK_TARGET_PASSWORD")
+
+	if host == "" || user == "" || password == "" {
+		fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_PASSWORD")
+		fmt.Println(
+		`Usage:
+		- SDK_TARGET_HOST				// Managment hostname or IP of array
```
rkumpf: *Management

pkg/service/volume_service_test.go
```diff
@@ -37,23 +37,60 @@ type VolumeServiceTestSuite struct {
 	tenantFolderService             *FolderService
 }
 
+func (suite *VolumeServiceTestSuite) checkEnvironmentVariableExists() {
```
rkumpf: This function seems redundant with the config() function, which should be called during the setup before the tests are run.  Can we remove it?

pkg/service/volume_service_test.go
```diff
@@ -125,16 +162,24 @@ func (suite *VolumeServiceTestSuite) TearDownTest() {
...
 	// Initialize volume attributes
 	var sizeField int64 = 5120
 	descriptionField := "This volume was created as part of a unit test"
 	var limitIopsField int64 = 256
 	var limitMbpsField int64 = 1
 
-	// TODO: replace <foldername> with your folder's name (owned by both admin and tenant)
-	// to make the test work
-	folder, _ := suite.nonTenantFolderService.GetFolderByName("<foldername>")
+	folder, _ := suite.nonTenantFolderService.GetFolderByName(os.Getenv("SDK_TARGET_TENANT_FOLDER"))
```
rkumpf: I suggest adding tenantFolder to the VolumeServiceTestSuite struct.  Then at lines 69 and 71, change tenantFolder to suite.tenantFolder.  Here, change os.Getenv("SDK_TARGET_TENANT_FOLDER") to suite.tenantFolder.

pkg/service/volume_service_test.go
```diff
@@ -37,23 +37,60 @@ type VolumeServiceTestSuite struct {
...
+
 /**
  * This function returns 2 group services:
  * The first return value is a group service for non-tenant
  * The second return value is a group service for tenant
  */
 func (suite *VolumeServiceTestSuite) config() (*NsGroupService, *NsGroupService) {
+	host := os.Getenv("SDK_TARGET_HOST")
+	user := os.Getenv("SDK_TARGET_USER")
+	password := os.Getenv("SDK_TARGET_USER_PASSWORD")
```
rkumpf: *SDK_TARGET_PASSWORD, for consistency with the examples.  Same thing in group_mgmt_client_test.go.

pkg/service/volume_service_test.go
```diff
@@ -125,16 +162,24 @@ func (suite *VolumeServiceTestSuite) TearDownTest() {
 }
 
 func (suite *VolumeServiceTestSuite) getDefaultVolumeOptions() *nimbleos.Volume {
+	folderId := os.Getenv("SDK_TARGET_TENANT_FOLDER")
```
rkumpf: We already retrieved and checked this variable in the config() function.  Can we remove this section?

pkg/client/group_mgmt_client_test.go
```diff
@@ -18,10 +19,16 @@ var (
 )
 
 func TestNewClient(t *testing.T) {
+	if (os.Getenv("SDK_TARGET_HOST") == "" ||
+		os.Getenv("SDK_TARGET_USER") == "" ||
+		os.Getenv("SDK_TARGET_USER_PASSWORD") == "") {
+			fmt.Println("ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_USER_PASSWORD");
+			os.Exit(1)
+		}
```
rkumpf: Align with "if".

pkg/client/group_mgmt_client_test.go
```diff
@@ -18,10 +19,20 @@ var (
...
+		fmt.Println("SDK_TARGET_HOST - Management hostname or IP of array")
+		fmt.Println("SDK_TARGET_USER - User (non-tenant) username")
+		fmt.Println("SDK_TARGET_PASSWORD - User (non-tenant) password")
+		os.Exit(1)
+	}
 
 	// Create client
 	var err error
-	client, err = NewClient("x.x.x.x", "xxx", "xxx", "v1", true, false)
+	client, err = NewClient(os.Getenv("SDK_TARGET_HOST"), os.Getenv("SDK_TARGET_USER"), os.Getenv("SDK_TARGET_USER_PASSWORD"), "v1", true, false)
```
rkumpf: s/SDK_TARGET_USER_PASSWORD/SDK_TARGET_PASSWORD/g

examples/filters.go
```diff
@@ -42,4 +56,8 @@ func main() {
 	//GetVolumes
 	volumes, err := volSvc.GetVolumes(sfilter)
 	fmt.Printf("%v", volumes)
+
+	if err != nil {
+		fmt.Printf("NewGroupService(): Unable to get volumes, err: %v", err.Error())
```
rkumpf: Add trailing \n.  Also at line 58 and elsewhere.

pkg/service/volume_service_test.go
```diff
@@ -43,17 +44,36 @@ type VolumeServiceTestSuite struct {
...
+	host := os.Getenv("SDK_TARGET_HOST")
+	user := os.Getenv("SDK_TARGET_USER")
+	password := os.Getenv("SDK_TARGET_PASSWORD")
+	tenant := os.Getenv("SDK_TARGET_TENANT_USER")
+	tenantPassword := os.Getenv("SDK_TARGET_TENANT_PASSWORD")
+	tenantFolder := os.Getenv("SDK_TARGET_TENANT_FOLDER")
+
+	if host == "" || user == "" || password == "" || tenant == "" || tenantPassword == "" || tenantFolder == "" {
+		suite.T().Error(
+			`ERROR: Missing one of these environment variables: SDK_TARGET_HOST, SDK_TARGET_USER, SDK_TARGET_USER_PASSWORD, SDK_TARGET_TENANT_USER, SDK_TARGET_TENANT_PASSWORD, SDK_TARGET_TENANT_FOLDER
```
rkumpf: SDK_TARGET_PASSWORD

rkumpf (approved):

rkumpf (approved):
Thanks for the updates!

PR #56, opened 2021-09-15
Add option to provide a custom tls.Config to allow control over the tls connection

jwatson-gcu:
## This PR proposes two things:
1) Adding a `service.ServiceOption` to provide a `tls.Config` to `NewNimbleGroupService()` which is in turn used by `resty.Client`
    
    The current behavior hard codes the tls connection to skip certificate verification, which is something that is not allowed in our security environment.
```
	restyClient.SetTLSClientConfig(&tls.Config{
		InsecureSkipVerify: true,
	})
```

2)  When using a **Tenant User**, certificates are **fully validated**.

Signed-off-by: Joseph Watson <jwatson@genisyscu.org>

### DIFF
--- a/pkg/client/group_mgmt_client.go
+++ b/pkg/client/group_mgmt_client.go
@@ -62,12 +62,16 @@ type Argument struct {
 	JobId string `json:"job_id,omitempty"`
 }
 
-func newGroupMgmtClient(ipAddress, username, password, apiVersion string, waitOnJobs, isTenant bool) *GroupMgmtClient {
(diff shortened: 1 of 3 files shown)


jwatson-gcu:
Example usage:
```
	nimbleClient, err := service.NewNimbleGroupService(
		service.WithHost(config.Host),
		service.WithUser(config.User),
		service.WithPassword(config.Password),
		service.WithTLSConfig(tlsConfig),
	)
```

